PR #12, opened 2015-03-06
Wercker

blisteringherb:
Adding the Wercker and Capistrano configs in place to deploy.

@angaither @jonuy


angaither:
looks amaaazing

PR #35, opened 2015-06-26
fixing bug issue #18

### LINKED ISSUE #18: Local admin user not seeded with correct perms
The local admin user `test@dosomething.org` is not seeded with the correct permissions.

axton21:
Fixes #18 where id is different from northstar.

For review: @angaither


angaither:
:+1: great, merge away!

Issue #30, opened 2015-06-25
Login page styles


angaither:
this should be both a modal or it's own page and use the form components 
http://neue.dosomething.org/#forms-button
http://neue.dosomething.org/#forms-form-item

paulhan221:
closed with #38

paulhan221:
closed with #38

PR #38, opened 2015-06-30
added modal styling to login

axton21:
Fixes issue #30
Using DS-modal
@angaither 
@DFurnes


npm-debug.log
```diff
@@ -0,0 +1,20 @@
+0 info it worked if it ends with ok
```
angaither: this file can be added to git ignore

app/controllers/SessionsController.php
```diff
@@ -34,7 +34,8 @@ public function store()
       return Redirect::route('users.index');
 
     } catch (Exception $e) {
-       return Redirect::route('login')->with('flash_message', ['class' => 'alert alert-danger', 'text' => 'Login failed'])->withInput();
+      $input['autoOpenModal'] = 'true';
```
DFurnes: This might make more sense as flash data, something like `->with('modal', true)`, rather than mixing it in with input.

app/views/sessions/create.blade.php
```diff
@@ -1,24 +1,23 @@
...
-    <h2 class="form-signin-heading">Please sign in</h2>
-    <div class="form-group">
+    <div class="form-item {{ 'modifier_class' }} -padded">
       {{ Form::label('email', 'Email', array('class' => 'sr-only')) }}
-      {{ Form::text('email', NULL, array('class' => 'form-control', 'placeholder' => 'Email Address')) }}
+      {{ Form::text('email', NULL, array('class' => 'text-field', 'placeholder' => 'Email Address')) }}
     </div>
-
-    <div class="form-group">
+    <div class="form-item {{ 'modifier_class' }} -padded">
```
DFurnes: You can remove `{{ 'modifier_class' }}` on both of these. That's a placeholder in the pattern library for adding modifiers like `-padded`.

app/views/layout/master.blade.php
```diff
@@ -5,29 +5,36 @@
...
-
 
-  @if (Session::has('flash_message'))
-    <div class="flash-message {{ Session::get('flash_message')['class'] }}">
-      <em>{{ Session::get('flash_message')['text'] }}</em>
+<body class="modernizr-no-js">
+  <div class="chrome">
+    @include('layout.nav')
+    @if (Session::has('flash_message'))
+      @if (Session::get('flash_message')['text'] === "Login Failed")
```
DFurnes: Hmm, this feels fragile to me. Not necessary to change for merge, but perhaps think about refactoring this somehow in the future?

DFurnes: Perhaps `Session::has('trigger_modal')` for the conditional and `window.DSModal.open($({{ Session::get('trigger_modal') }}))` to open the modal, so that it's useable in other scenarios?

DFurnes:
:shipit:!

axton21:
I saw Mice!

PR #39, opened 2015-07-01
User index

### LINKED ISSUE #31: Users listing page
Use [waypoints](http://neue.dosomething.org/#waypoints) to style the pagination 
And the table grid thing (not yet built in neue) for the user listing
[Inline form styles](http://neue.dosomething.org/#forms-form-item) for the search functionality

axton21:
Fixes #31 
@angaither 
@DFurnes


DFurnes:
Looks good! :rocket: 

One future improvement could be moving the pagination partial into a [custom presenter](http://laravel.com/docs/4.2/pagination#custom-presenters) so that it can be used for multiple types of data, rather than just for users.

Issue #33, opened 2015-06-25
User detail page


angaither:
this might be a new addition to neue, since there isn't a lot form patterns in neue yet. 

@DFurnes can provide more detail here

angaither:
Campaign info will be displayed on this page as well, and it can follow the [duo-gallery with figures inside it pattern](http://neue.dosomething.org/#gallery-duo)

paulhan221:
closed with #42

paulhan221:
closed with #42

PR #42, opened 2015-07-06
User show page

paulhan221:
Fixes issue #33 
@angaither 
@DFurnes

### DIFF
--- a/app/Aurora/helpers.php
+++ b/app/Aurora/helpers.php
@@ -1,10 +1,10 @@
 <?php
 // Global helper functions.
 function autoOpenModal(){
-echo
-'<script>
-$(document).ready(function () {
-window.DSModal.open($("#signin-modal"));
-});
-</script>';
+	return
+	'<script>
+	$(document).ready(function () {
+	window.DSModal.open($("#signin-modal"));
+	});
+	</script>';
 }
--- a/app/Aurora/helpers.php
+++ b/app/Aurora/helpers.php
@@ -8,10 +8,3 @@ window.DSModal.open($("#signin-modal"));
 });
 </script>';
 }
-
-function keySantizer($key){
-$key = str_replace("_"," ", $key);
-$key = preg_replace('/^\s/',"", $key);
-$key = ucfirst(preg_replace('/addr /',"", $key));
-return $key;
-}
--- a/app/views/layout/nav.blade.php
+++ b/app/views/layout/nav.blade.php
@@ -3,7 +3,7 @@
   <div class="navigation__menu">
   <ul class="navigation__primary">
     <li>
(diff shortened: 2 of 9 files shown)


app/Aurora/helpers.php
```diff
@@ -1,10 +1,17 @@
 <?php
 // Global helper functions.
 function autoOpenModal(){
-	echo
-		'<script>
-			$(document).ready(function () {
-				window.DSModal.open($("#signin-modal"));
-			});
-		</script>';
+echo
```
DFurnes: I would recommend returning this as a string, which will then print from wherever you call the function in your blade templates.

app/Aurora/helpers.php
```diff
@@ -1,10 +1,17 @@
...
+'<script>
+$(document).ready(function () {
+window.DSModal.open($("#signin-modal"));
+});
+</script>';
+}
+
+function keySantizer($key){
```
DFurnes: Where is this function used?

public/assets/custom-neue.css
```diff
@@ -53,3 +53,7 @@
 .pagination-buttons {
 	display:inline;
 }
+
+.account-info {
+  margin: 2% 0;
```
DFurnes: This is probably something you can do with built-in Neue container classes. If not, let's take some time to set you guys up writing SCSS so that you can use Neue's variables and layout helpers.

app/Aurora/helpers.php
```diff
@@ -1,10 +1,10 @@
...
 // Global helper functions.
 function autoOpenModal(){
-	echo
-		'<script>
-			$(document).ready(function () {
-				window.DSModal.open($("#signin-modal"));
-			});
-		</script>';
-}
+echo
```
DFurnes: You can use `return '<script> ...` here instead of `echo '<script> ...`. When you call the function from within a Blade tag, it'll get placed into the page (although you'll have to use `{!!` and `!!}` because it shouldn't be escaped).

app/views/layout/nav.blade.php
```diff
@@ -1,9 +1,9 @@
 <nav class="navigation">
-  <a class="navigation__logo" href=""><span>DoSomething.org</span></a>
+  <a class="navigation__logo" href="/"><span>DoSomething.org</span></a>
   <div class="navigation__menu">
   <ul class="navigation__primary">
     <li>
-      <h1 class="heading -hero">Aurora</h1>
+      <a href="."><h1 class="heading -hero">Aurora</h1></a>
```
DFurnes: Is the additional link tag here a mistake? I'm not sure that `href="."` will link anywhere. :grimacing:

DFurnes:
:shipit:

PR #53, opened 2015-07-07
added nav subtitle

paulhan221:


public/assets/custom-neue.css
```diff
@@ -3,6 +3,10 @@
 .navigation {
     background-color: #4e2b63;
 }
+
+.navigation__subtitle {
+  color: white;
```
DFurnes: This shouldn't be necessary to override.

public/assets/custom-neue.css
```diff
@@ -3,6 +3,10 @@
 .navigation {
```
DFurnes: You can use the [header](http://neue.dosomething.org/#header) pattern (alongside the navigation pattern's floating modifier, which you're already using) to do this without an override.

public/assets/custom-neue.css
paulhan221: it shows up as the same color as the background otherwise
it seems to be the default color unless specified

angaither:
:+1:

DFurnes:
:shipit:!

Issue #49, opened 2015-07-07
Keys pages


angaither:
- [ ]  add new key page should be in form styles
- [ ] keys listing page should display app_id and app_key as one unit (maybe gallery-duo type)
@DFurnes what do you think

paulhan221:
closed with #56

PR #56, opened 2015-07-08
Keys

paulhan221:
References #49 
@DFurnes 
@angaither


app/views/keys/index.blade.php
```diff
@@ -1,20 +1,23 @@
...
-     <dt class="control-label col-sm-2"><strong> API Key </strong> </dt>
-     <dl> {{ $key['api_key'] }} </dl>
-    </div>
-  @endforeach
-@endif
-
-@stop
+<div class="container -padded">
+	<div class="container__block">
+		<a href="{{ URL::Route('keys.create') }}">New App</a>
```
DFurnes: You can use the `route()` helper method here, which is a bit cleaner to read.

app/views/keys/index.blade.php
```diff
@@ -1,20 +1,23 @@
...
+	</div>
+		<h2> Listing of all Northstar API keys</h2>
+		@if ($keys)
+			<ul class="gallery -duo">
+			  @foreach($keys as $key)
```
DFurnes: It might be nice to use a `@forelse` block here so you can output "No API keys" if there aren't any keys, rather than just not printing anything at all.

DFurnes:
Two comments, everything else looks good to me! :ship:

Issue #48, opened 2015-07-07
User profile page 


angaither:
- [ ] take a look at the profile page in drupal to show user profile information 
- [ ] print campaign info box without variable types
- [ ] style the mobile numbers to a standard format
- [ ] make admin button should be red and not on the top as a primary action

PR #57, opened 2015-07-08
sanitize phone number

axton21:
Reference #48 
@DFurnes 
@angaither


app/Aurora/helpers.php
```diff
@@ -8,3 +8,9 @@ function autoOpenModal(){
 	});
 	</script>';
 }
+
+function sanitizePhoneNumber($number){
+	$number = preg_replace('/[^0-9]/', '', $number);
+	$number = preg_replace("/([0-9]{3})([0-9]{3})([0-9]{4})/", "$1-$2-$3", $number);
```
DFurnes: This would run into problems with 11 digit international numbers, e.g. `(xxx) xxxx xxxx` from the UK.

aaronschachter: Don't mind me jumping in but we were just chatting about international validation here: https://github.com/DoSomething/product/issues/32

aaronschachter: Specifically potentially looking into https://github.com/googlei18n/libphonenumber.

Also a super vintage issue here  https://github.com/DoSomething/dosomething/issues/65 to look into https://github.com/respect/validation

angaither: I think the validation thing is something we should for sure try to standardize across our apps, but this issue is just for displaying the phone numbers we already have in a better format.

aaronschachter: Got it.
i'll be leaving now

axton21:
References #48 
@DFurnes 
@angaither 

It works, but some fake data phone numbers are just not relevant at all.
Thanks @aaronschachter! Very useful google library.

angaither:
looks like you need to rebase this branch before merging

app/Aurora/helpers.php
```diff
@@ -8,3 +8,14 @@ function autoOpenModal(){
...
 }
+
+//https://github.com/giggsey/libphonenumber-for-php
+function sanitizePhoneNumber($number){
+	$phoneUtil = \libphonenumber\PhoneNumberUtil::getInstance();
+	try {
+		$number = $phoneUtil->parse($number, "US");
+		return $phoneUtil->format($number, \libphonenumber\PhoneNumberFormat::INTERNATIONAL);
+	} catch (\libphonenumber\NumberParseException $e) {
+    return $e;
```
DFurnes: Unless these exceptions are intended to be user-facing, this might make more sense to save to the log & then just return the unformatted number.

composer.json
```diff
@@ -6,7 +6,8 @@
   "type": "project",
   "require": {
     "laravel/framework": "4.2.*",
-    "guzzlehttp/guzzle": "5.1.*"
+    "guzzlehttp/guzzle": "5.1.*",
+     "giggsey/libphonenumber-for-php": "~7.0"
```
DFurnes: Indentation looks a bit off here.

app/Aurora/helpers.php
```diff
@@ -8,3 +8,14 @@ function autoOpenModal(){
 	});
 	</script>';
 }
+
+//https://github.com/giggsey/libphonenumber-for-php
+function sanitizePhoneNumber($number){
+	$phoneUtil = \libphonenumber\PhoneNumberUtil::getInstance();
+	try {
+		$number = $phoneUtil->parse($number, "US");
```
angaither: we will be using this with international folks, so maybe we can pass in the country code from the user doc to parse the number based on the country we have associated with that user

DFurnes: Oh, +1 +1 +1! That'd be smart.

axton21:
@DFurnes @DFurnes

app/Aurora/helpers.php
```diff
@@ -8,3 +8,19 @@ function autoOpenModal(){
...
+
+//https://github.com/giggsey/libphonenumber-for-php
+function sanitizePhoneNumber($number, $countryName){
+	if(!$countryName){
+		$countryName = "US";
+		//checking if country code is found, if not will just be US as default
+	}
+	$phoneUtil = \libphonenumber\PhoneNumberUtil::getInstance();
+	try {
+		$number = $phoneUtil->parse($number, $countryName);
```
DFurnes: I'd suggest renaming this to `$formattedNumber` since otherwise it is not clear you're overwriting the given parameter.

app/Aurora/helpers.php
```diff
@@ -8,3 +8,19 @@ function autoOpenModal(){
 	});
 	</script>';
 }
+
+//https://github.com/giggsey/libphonenumber-for-php
+function sanitizePhoneNumber($number, $countryName){
+	if(!$countryName){
```
DFurnes: This can be handled using default parameters:

```php
function sanitizePhoneNumber($number, $countryName = 'US'){
  // ...
}
```

DFurnes:
Two comments. Also looks like you might need to fix some merge conflicts! :grimacing:

app/Aurora/helpers.php
axton21: Good call, thanks!

Issue #27, opened 2015-06-23
Add campaign data for signups


angaither:
Call the drupal api to get the campaign title and header photo for the campaign to display on the user detail page.

This will be themed with the gallery duo with figures style

PR #58, opened 2015-07-10
Campaigns

paulhan221:
References Issue #27 
@angaither 
@DFurnes 

Sorry if this pull request is super long, its because we realized we needed to move container & wrapper out of the master page and to everywhere else and then we moved the header from the master page to everywhere else.


app/controllers/UsersController.php
```diff
@@ -55,13 +55,22 @@ public function store()
    */
   public function show($id)
   {
+    $campaigns = [];
     $user = Session::get('user');
     if (!$user) {
       $northstar = new Aurora\Services\Northstar\NorthstarAPI;
       $user = $northstar->getUser('_id', $id);
       $aurora_user = User::where('_id', $id)->first();
+      $drupal = new Aurora\Services\Drupal\DrupalAPI;
```
DFurnes: This should be injected in the controller's constructor:

```php
  use Aurora\Services\Drupal\DrupalAPI;

  public function __construct(DrupalAPI $drupal) {
     $this->beforeFilter('auth');
     $this->beforeFilter('role:admin');

     $this->drupal = $drupal;
   }

  // ...

  if (!empty($user['campaigns'])){
       foreach($user['campaigns'] as $campaign){
           array_push($campaigns, $this->drupal->getCampaign($campaign['drupal_id']));
       }
  }

```

It doesn't make a huge difference at the moment, but it's important if we add tests to these pages.

app/controllers/UsersController.php
```diff
@@ -55,13 +55,22 @@ public function store()
...
+
+      if (!empty($user['campaigns'])){
+        foreach($user['campaigns'] as $campaign){
+           array_push($campaigns, $drupal->getCampaign($campaign['drupal_id']));
+        }
+      }
     }
-    return View::make('users.show')->with(compact('user', 'aurora_user'));
+    $drupal = new Aurora\Services\Drupal\DrupalAPI;
```
DFurnes: This looks like a mistake.

app/views/keys/create.blade.php
```diff
@@ -1,21 +1,28 @@
 @extends('layout.master')
 
 @section('main_content')
+<header class="header" role="banner">
```
DFurnes: It might be nice to make a shared `header.blade.tpl` partial that accepts a `title` and `subtitle` variable, so you don't have to copy this markup into each template.

DFurnes:
Couple suggestions – looking good so far.

app/Aurora/Services/Drupal.php
```diff
@@ -0,0 +1,25 @@
+<?php namespace Aurora\Services\Drupal;
+
+class DrupalAPI {
+
+  protected $client;
+
+  public function __construct()
+  {
+    $base_url = "http://staging.beta.dosomething.org/api/v1/";
```
angaither: we will want to set these as variables and not hard code this so it can be different per environment

app/Aurora/Services/Drupal.php
```diff
@@ -0,0 +1,25 @@
...
+    $client = new \GuzzleHttp\Client(['base_url' => $base_url]);
+
+    $this->client = $client;
+  }
+
+
+  public function getCampaign($cid)
+  {
+
+    $response = $this->client->get('campaigns/' . $cid);
```
angaither: let's just call this `id` instead of `cid` (we've been trying to stick to that standard across our apps)

app/controllers/UsersController.php
```diff
@@ -55,13 +55,22 @@ public function store()
...
     $user = Session::get('user');
     if (!$user) {
       $northstar = new Aurora\Services\Northstar\NorthstarAPI;
       $user = $northstar->getUser('_id', $id);
       $aurora_user = User::where('_id', $id)->first();
+      $drupal = new Aurora\Services\Drupal\DrupalAPI;
+
+      if (!empty($user['campaigns'])){
+        foreach($user['campaigns'] as $campaign){
+           array_push($campaigns, $drupal->getCampaign($campaign['drupal_id']));
```
angaither: same here, use `id`

app/Aurora/Services/Drupal.php
paulhan221: We actually had a question about this. We were wondering how the Northstar API is working when the $base_url for it is in the .env file, but we don't have an .env anywhere?

app/controllers/UsersController.php
paulhan221: Should we inject all the $northstar initializations as well?

paulhan221: should we do this for all the $northstar initializations as well?

DFurnes: Yes please! Didn't notice the other one in there.

app/Aurora/Services/Drupal.php
```diff
@@ -0,0 +1,25 @@
...
+
+class DrupalAPI {
+
+  protected $client;
+
+  public function __construct()
+  {
+    $base_url = \Config::get('services.drupal.url');
+
+    $client = new \GuzzleHttp\Client(['base_url' => $base_url]);
```
DFurnes: Not necessary for merge, but it might be cleaner to import this class with a `use` statement at the top of the file:

```php
<?php

use GuzzleHttp\Client;

// ...

$client = new Client(['base_url' => $base_url]);

File: app/config/local/services.php [hunk 8]
```diff
@@ -20,4 +20,6 @@
       'app_id'  => '456',
       'api_key' => 'abc4324'
     ),
+    'drupal' => array(
+      'url' => "http://staging.beta.dosomething.org/api/v1/")
```
DFurnes: Small code style quibble: you should put the closing parenthesis here on a newline (as in the `northstar` entry above).

app/Aurora/Services/Drupal.php
angaither: :+1: @DFurnes

DFurnes:
Two tiny comments, but everything else looks good to ship! :ship:

Issue #34, opened 2015-06-25
Reportback information


angaither:
Call the drupal api to pull in more information about the reportback

- [ ] reportback photo
- [ ] reportback quantity 
- [ ] campaign title

PR #59, opened 2015-07-13
Reportback item

paulhan221:
References #34
@DFurnes 
@angaither 

We had a question about whether or not we should display all reportback-items for each campaign or just the first one? The staging.beta site only displays the first reportback-item so we went with that.

Also, should we also display the caption and "why participated" in the reportbacks or just the 3 things you referenced in the issue? We only displayed the 3 mentioned in the issue.


angaither:
Just the first one is fine, and this content is good for now!
:+1:

PR #63, opened 2015-07-14
Fix user attributes

### LINKED ISSUE #62: User attributes aren't displaying properly
Couple of things I noticed when migrating real user data in

- Address field displays a comma, even if there isn't an address
- Drupal_id doesn't return at all, but from northstar, the users have a value there
- Last name: displays even if the user doesn't have a last name

axton21:
Fixes #62 
@angaither 
@DFurnes

### DIFF
--- a/app/views/users/partials/details.blade.php
+++ b/app/views/users/partials/details.blade.php
@@ -3,18 +3,17 @@
 		<article class="figure -left">
 			<h3>Account Info</h3>
 			<dl class="profile-settings">
-				<dt>Id:</dt><dd>{{ $user['_id'] or '' }}</dd>
-				{{ isset($user['drupal_id']) ? ('<dt>Drupal Id:</dt><dd>'.$user['drupal_id']) : "" }}
-				{{ isset($user['firt_name']) ? ('<dt>First Name:</dt><dd>'.$user['firt_name']) : "" }}
-				{{ isset($user['last_name']) ? ('<dt>Last Name:</dt><dd>'.$user['last_name']) : "" }}
-				{{ isset($user['email']) ? ('<dt>Email:</dt><dd>'.$user['email']) : "" }}
-				{{ isset($user['mobile']) ? ('<dt>Mobile:</dt><dd>'.$user['mobile']) : "" }}
-				{{ isset($user['birthdate']) ? ('<dt>Birthday:</dt><dd>'.$user['birthdate']) : "" }}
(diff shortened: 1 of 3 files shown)


app/views/users/partials/details.blade.php
```diff
@@ -4,13 +4,17 @@
...
 			<dl class="profile-settings">
 				<dt>Id:</dt><dd>{{ $user['_id'] or '' }}</dd>
-				<dt>First Name:</dt><dd>{{ $user['first_name'] or '' }}</dd>
-				<dt>Last Name:</dt><dd>{{ $user['last_name'] or '' }}</dd>
-				<dt>Email:</dt><dd>{{ $user['email'] or '' }}</dd>
-				<dt>Mobile:</dt><dd>{{ $user['mobile'] or '' }}</dd>
-				<dt>Birthday:</dt><dd>{{ $user['birthdate'] or '' }}</dd>
-				<dt>Address:</dt><dd>{{ $user['addr_street1'] or ''}} {{ $user['addr_street2'] or ''}} {{ $user['add_city'] or ''}}, {{ $user['addr_state'] or ''}} {{ $user['addr_zip'] or ''}}</dd>
-				<dt>Country:</dt><dd>{{ $user['country'] or '' }}</dd>
+				{{ isset($user['drupal_id']) ? ('<dt>Drupal Id:</dt><dd>'.$user['drupal_id']) : "" }}
```
DFurnes: These are starting to look a bit messy. Perhaps this could be good to create a helper function for? Also, since these are user inputted content it would be good to sanitize with `e()`.
```
{{ profile_field('Drupal ID', $user['drupal_id']) }}
```

DFurnes: (Or, if you decide against making a helper function, you can just wrap the user data in `{{{ }}}` which is Laravel 4's escaped blade syntax.)

axton21:
@DFurnes @angaither

DFurnes:
Looks like you've got some merge conflicts. :fearful: After that, good to merge! :soon:

axton21:
I just fix conflicts, @DFurnes is that because we merge PR after this commits? Thanks!

DFurnes:
Merge conflicts happen when you make changes to the same line in two separate branches that Git can't figure out how to merge. Not sure what happened in this particular case, but luckily it's usually pretty easy to fix. :)

DFurnes:
:ship: away!

axton21:
Yass~~

PR #64, opened 2015-07-15
Sanitize phone number

axton21:
Reference #48 
@DFurnes @angaither 

Seems like github has been acting a little weird that I decided to close the other one and make same new pull request, the last commit was linked to an older commits.


angaither:
:+1:

PR #70, opened 2015-07-16
Login page

### LINKED ISSUE #67: Style the `login` page
the modal login is styled, but when you go to `/login` it's an unstyled page

axton21:
Fixes #67

@DFurnes @angaither 

So in this PR we restyle the login page template and did some change  and renaming to DS modal where we used to put it in master and its unecessary to have signin-modal in every page.


DFurnes:
Seems fine to me. :boat:

Issue #10, opened 2015-02-12
Mobile Commons integration - *Priority 3*


angaither:
Helpful?

mshmsh5000:
Yes

mshmsh5000:
To be clear, the first iteration of this will involve directly accessing the Mobile Commons API to display related user data.

jonuy:
Info on the Mobile Commons endpoint for getting user info: https://mobilecommons.zendesk.com/hc/en-us/articles/202052534-REST-API#ProfileSummary

PR #71, opened 2015-07-16
Mobile commons

paulhan221:
References issue #10 
@angaither 
@DFurnes


app/controllers/UsersController.php
```diff
@@ -137,5 +144,15 @@ public function adminCreate($user_id)
     return Redirect::back()->with('flash_message', ['class' => 'alert alert-success', 'text' => 'The more admins the merrier.']);
   }
 
+  public function mobileCommonsMessages($id)
+  {
+    $user = $this->northstar->getUser('_id', $id);
+
+    // $mc_messages = $this->mobileCommons->userMessages($user['mobile']);
+    $mc_messages = $this->mobileCommons->userMessages('3473276087');
```
angaither: think this was left in by mistake

paulhan221: O_O whoops

app/Aurora/helpers.php
```diff
@@ -8,3 +8,12 @@ function autoOpenModal(){
 	});
 	</script>';
 }
+
+
+function timeFormatter($date){
```
DFurnes: Laravel coding conventions use "snake case" for helper methods, so this should be `time_formatter`.

app/views/users/mobile-commons-messages.blade.php
```diff
@@ -0,0 +1,51 @@
...
+
+@section('main_content')
+
+@include('layout.header', ['header' => 'Mobile Commons Messages', 'subtitle' => 'Backlog of all texts'])
+
+<div class ="container -padded">
+  <div class="wrapper">
+	  <div class="container__block">
+	  	@forelse($mc_messages as $message)
+	  		@if ($message['@attributes']['status'] == "sent")
```
DFurnes: I would recommend figuring out this modifer class (`sent` vs `received` vs `failed`) using a helper method, so you don't need to repeat the opening tag in each conditional, so maybe something like this:

```html
<div class="message {{ message_state_class($message) }}">
```

DFurnes: We also use `is-` or `has-` prefixes for state modifier classes, to help avoid naming conflicts. So in this case: `is-sent`, `is-received` or `is-failed`.

public/assets/custom-neue.css
```diff
@@ -74,3 +74,30 @@
...
+.message {
+  border-radius: 5px;
+  float: right;
+  width: 100%;
+  position: relative;
+  margin: 10px 0;
+  padding-left: 2%;
+}
+
+.sent {
```
DFurnes: These should use `is-` prefix, and be namespace to the `.message` class to prevent collisions, so for example: 

```css
.message.is-sent {
  /* ... */
}

.message.is-received {
  /* ... */
}
```

public/assets/custom-neue.css
```diff
@@ -74,3 +74,30 @@
 .profile-settings dd{
   float: left;
 }
+
+.message {
```
DFurnes: I would recommend renaming this to `mobilecommons-message` (or something similar), since it's easy to get confused with our existing [`.messages`](http://neue.dosomething.org/#messages) pattern with this naming.

public/assets/custom-neue.css
```diff
@@ -74,3 +74,30 @@
...
+  border-radius: 5px;
+  float: right;
+  width: 100%;
+  position: relative;
+  margin: 10px 0;
```
DFurnes: We use multiples of `24px` for spacing, so this should be `12px`.

app/Aurora/Services/MobileCommons.php
```diff
@@ -0,0 +1,55 @@
...
+  public function userProfile($mobile)
+  {
+
+    $response = $this->client->get('?phone_number=' . $mobile . '&include_messages=true');
+
+    $xml = $response->xml();
+
+    $json = json_encode($xml);
+
+    $array = json_decode($json, TRUE);
```
angaither: I'm not sure the format you're looking for here is, but does [xml_parse_into_struct](http://php.net/manual/en/function.xml-parse-into-struct.php) work?

app/Aurora/Services/MobileCommons.php
```diff
@@ -0,0 +1,55 @@
...
+
+    return $array['profile'];
+  }
+
+  public function userMessages($mobile)
+  {
+
+    $response = $this->client->get('?phone_number=' . $mobile . '&include_messages=true');
```
angaither: can this be broken out into a function if we are sending the same request to get messages in two places?

app/Aurora/helpers.php
angaither: style, curly braces of functions on the next line

public/assets/custom-neue.css
```diff
@@ -74,3 +74,30 @@
 .profile-settings dd{
   float: left;
 }
+
+.message {
+  border-radius: 5px;
```
DFurnes: We have [preset `border-radius` sizes](https://github.com/DoSomething/neue/blob/8c05988ac995dc1cb5f1129e4d41cdad567bc97f/scss/_utilities/_variables.scss#L94-L95) that we use in our styles. Since these elements are not supposed to be clickable, they should probably be using the `2px` radius.

angaither:
@paulhan221 @axton21 can you guys attach some screenshots of where/how the messages are attached (as long as the data is fake.. if it's real throw them in slack)
I am curious to see how/where they are attached to the user display

angaither:
Also, it looks like this PR needs to be rebased?

DFurnes:
I made a couple of nitpicky style comments – it might be worth taking a little time to import Neue [via `neue.scss`](https://github.com/DoSomething/neue/blob/8c05988ac995dc1cb5f1129e4d41cdad567bc97f/scss/neue.scss) so you can take advantage of all of it's variables and mixins. This should be pretty straightforward using Laravel Elixir's built-in `mix.sass()` function.

paulhan221:
image

The status for the messages are all "failed permanently" bc its not real data.

paulhan221:
image

for statuses that are received, and sent

paulhan221:
I'm also using my personal account for mobile commons to authenticate the requests, should we use another account?

app/Aurora/Services/MobileCommons.php
paulhan221: We're trying to get it into JSON format, I'm not sure if this will work, there are quite a few nested attributes.

angaither:
yeah, we should use another account for authentication

paulhan221:
@DFurnes 
@angaither

app/views/users/partials/mobile-commons.blade.php
```diff
@@ -0,0 +1,20 @@
...
+			@if (!empty($mc_profile))
+				<h3>Account Info</h3>
+				<dl class="profile-settings">
+					<dt>Mobile Commons Id:</dt><dd>{{{ $mc_profile['@attributes']['id'] or '' }}}</dd>
+					<dt>Status:</dt><dd>{{{ $mc_profile['status'] or '' }}}</dd>
+					{{ isset($mc_profile['created_at']) ? ('<dt>Signed Up On:</dt><dd>' . e(time_formatter($mc_profile['created_at']))) : "" }}
+					{{ isset($mc_profile['opted_out_at']) ? ('<dt>Opted Out On:</dt><dd>' . e(time_formatter($mc_profile['opted_out_at']))) : "" }}
+					{{ isset($mc_profile['opted_out_source']) ? ('<dt>Opted Out Source:</dt><dd>' . e($mc_profile['opted_out_source'])) : "" }}
+					<dt><a href="https://secure.mcommons.com/profiles/{{{ $mc_profile['@attributes']['id'] }}}">View Mobile Commons Profile</a></dt>
+					<dt><a href="{{{ url('users/' . $user['_id'] . '/mobile-commons-messages') }}}">View Message Backlog</a></dt>
```
angaither: can use laravels helper [link_to](http://laravel.com/docs/4.2/helpers#urls)

DFurnes: I've actually gotten in the habit of just using `url()` everywhere in my code, since you need to use the ugly raw input tags `{!! !!}` to use `link_to` in Laravel 5.

paulhan221:
@DFurnes Hm maybe I'm just not using it right but no matter what combination I used, I wasn't able to figure out how to use url() instead of link_to for the mobile-commons-profile link.

DFurnes:
You were already using it!

paulhan221:
@DFurnes @angaither 
is this ready to merge?

DFurnes:
Yep, looks good to me! :shipit:

PR #76, opened 2015-07-20
removed underscore

axton21:


axton21:
@DFurnes

app/views/users/partials/index-table.blade.php
```diff
@@ -1,7 +1,7 @@
 <table id="user-table">
 	<thead>
 		<tr class="row table-header">
-			<th class="table-cell">_id</th>
+			<th class="table-cell">Id</th>
```
DFurnes: Would it make more sense to rename this "User ID"?

axton21:
sure...

axton21:
@DFurnes

DFurnes:
:boom: :shipit:

Issue #72, opened 2015-07-17
Display all northstar admins


angaither:
filter users to show me 'only admins' so you can add/remove people from the admin list

axton21:
@angaither do you mean aurora admin user in database?

PR #79, opened 2015-07-20
fixing page error for no admin found

axton21:
References #72
@angaither @DFurnes 

qa page errors out when there is no admin user found.

### DIFF
--- a/app/views/users/admin-index.blade.php
+++ b/app/views/users/admin-index.blade.php
@@ -6,11 +6,7 @@
 <div class="container -padded">
   <div class="wrapper">
     <div class="container__block">
-      @if(empty($users))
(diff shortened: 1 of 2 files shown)


app/views/users/admin-index.blade.php
```diff
@@ -6,7 +6,11 @@
 <div class="container -padded">
   <div class="wrapper">
     <div class="container__block">
-      @include('users.partials.index-table')
+      @if(empty($users))
```
DFurnes: This could also be handled by using a `@forelse` in the `index-table` blade template.

axton21:
@DFurnes Right, that makes alot more sense. Thanks for noticing that.

DFurnes:
:shipit:

PR #84, opened 2015-07-20
Using partial for login form

axton21:
@DFurnes @angaither 
So, using a partial form instead of having 2 forms for login.


angaither:
:+1: :ship: :shipit:

PR #85, opened 2015-07-20
conditional checking if campaigns doesnt have drupal_id

### LINKED ISSUE #81: User show fails if no campaigns
I'm getting an error on the first "test" user in prod

`cURL error 6: Could not resolve host: campaigns`
thinking maybe it's failing because the user doesn't have any campaigns, or the campaigns they are signed up for don't exist on drupal

axton21:
Fixes #81 @angaither @DFurnes @jonuy 

Talking about old northstar data that doesnt gives back drupal_id with @jonuy.
Hopefully this will solve the problem for now.

### DIFF
--- a/app/controllers/UsersController.php
+++ b/app/controllers/UsersController.php
@@ -67,9 +67,11 @@ class UsersController extends \BaseController {
       $aurora_user = User::where('_id', $id)->first();
       if (!empty($user['campaigns'])){
         foreach($user['campaigns'] as $campaign){
-          array_push($campaigns, $this->drupal->getCampaign($campaign['drupal_id']));
-          if (!empty($campaign['reportback_id'])) {
-            array_push($reportbacks, $this->drupal->getReportbacks($campaign['reportback_id']));
+          if (!empty($campaign['drupal_id'])) {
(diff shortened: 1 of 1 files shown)


app/controllers/UsersController.php
```diff
@@ -67,9 +67,11 @@ public function show($id)
       $aurora_user = User::where('_id', $id)->first();
       if (!empty($user['campaigns'])){
         foreach($user['campaigns'] as $campaign){
-          array_push($campaigns, $this->drupal->getCampaign($campaign['drupal_id']));
-          if (!empty($campaign['reportback_id'])) {
-            array_push($reportbacks, $this->drupal->getReportbacks($campaign['reportback_id']));
+          if (!empty($campaign['drupal_id'])) {
+            array_push($campaigns, $this->drupal->getCampaign($campaign['drupal_id']));
```
angaither: what if `getCampaign` fails? I think that may be the case for some of the test data

axton21: Yup, working on it =)
Need to make another filter in `drupal->getCampaign`

angaither: Okay, that can be a different PR!

angaither:
:shipit: :rocket:

PR #88, opened 2015-07-21
accounted for undefined index error in drupal functions

paulhan221:
@DFurnes 
@angaither

### DIFF
--- a/app/controllers/UsersController.php
+++ b/app/controllers/UsersController.php
@@ -79,8 +79,6 @@ class UsersController extends \BaseController {
       }
     }
     $mc_profile = $this->mobileCommons->userProfile($user['mobile']);
-
-    $mc_messages = $this->mobileCommons->userMessages($user['mobile']);
     
     return View::make('users.show')->with(compact('user', 'aurora_user', 'campaigns', 'reportbacks', 'mc_messages', 'mc_profile'));
   }
--- a/app/views/users/partials/mobile-commons.blade.php
+++ b/app/views/users/partials/mobile-commons.blade.php
@@ -7,8 +7,8 @@
 					<dt>Mobile Commons Id:</dt><dd>{{{ $mc_profile['@attributes']['id'] or '' }}}</dd>
 					<dt>Status:</dt><dd>{{{ $mc_profile['status'] or '' }}}</dd>
 					{{ isset($mc_profile['created_at']) ? ('<dt>Signed Up On:</dt><dd>' . e(time_formatter($mc_profile['created_at']))) : "" }}
-					{{ isset($mc_profile['opted_out_at']) ? ('<dt>Opted Out On:</dt><dd>' . e(time_formatter($mc_profile['opted_out_at']))) : "" }}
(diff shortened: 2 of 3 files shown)


angaither:
what does the page do when this fails, doesn't the user controller expect a response?

paulhan221:
image

When it fails nothing gets pushed into the array and an empty array gets returned, which is taken care of by the forelse statement....

Not sure if this is the best way to do this though =/

app/controllers/UsersController.php
```diff
@@ -79,8 +79,6 @@ public function show($id)
       }
     }
     $mc_profile = $this->mobileCommons->userProfile($user['mobile']);
-
-    $mc_messages = $this->mobileCommons->userMessages($user['mobile']);
```
angaither: is this meant to be removed? you're still passing it into the view below

app/views/users/partials/mobile-commons.blade.php
```diff
@@ -7,8 +7,8 @@
 					<dt>Mobile Commons Id:</dt><dd>{{{ $mc_profile['@attributes']['id'] or '' }}}</dd>
 					<dt>Status:</dt><dd>{{{ $mc_profile['status'] or '' }}}</dd>
 					{{ isset($mc_profile['created_at']) ? ('<dt>Signed Up On:</dt><dd>' . e(time_formatter($mc_profile['created_at']))) : "" }}
-					{{ isset($mc_profile['opted_out_at']) ? ('<dt>Opted Out On:</dt><dd>' . e(time_formatter($mc_profile['opted_out_at']))) : "" }}
-					{{ isset($mc_profile['opted_out_source']) ? ('<dt>Opted Out Source:</dt><dd>' . e($mc_profile['opted_out_source'])) : "" }}
+					{{ !empty($mc_profile['opted_out_at']) ? ('<dt>Opted Out On:</dt><dd>' . e(time_formatter($mc_profile['opted_out_at']))) : "" }}
+					{{ !empty($mc_profile['opted_out_source']) ? ('<dt>Opted Out Source:</dt><dd>' . e($mc_profile['opted_out_source'])) : "" }}
```
angaither: I think it's probably best to stick with `isset` since it's what is used across the the rest of the views in the app.
If possible, let's only set the vars that need to be
Maybe only push a value onto the array if a something is returned from the call to drupal in the code before this?

paulhan221: This is accounting for a unique case where the user opted out, and then opted back in. So when the user never opted out, the "opted_out_at" attribute doesn't exist. But when the user opts out and opts back in, the "opted_out_at" attribute exists, but only as an empty array. So when we use isset it errors out bc the "time_formatter" function is expecting a string, but an empty array is given. @angaither

angaither: oh, ew gross. Can you add that in as a comment for the ~future~

paulhan221:
Closing this pull request because it will conflict greatly with PR #90 
PR #90 will resolve all conflicts this pull request was accounting for and more

PR #90, opened 2015-07-22
UserController Show -- Refactoring

axton21:
@angaither @DFurnes 
Basically this PR refactored most userController show function which handles calling multiple API which is not good to do in controller.
The solution is we mostly create a northstarUser class to handle all the API calls related with that user.
Makes the code easier to read in controller and managable in NorthstarUser.php class.

@angaither 
This also fixing part of Issue #81 in Drupal.php for no campaigns from API as well which was closed in PR #85.

### DIFF
--- a/app/Aurora/Services/MobileCommons.php
+++ b/app/Aurora/Services/MobileCommons.php
@@ -24,27 +24,24 @@ class MobileCommonsAPI {
     $this->client = $client;
   }
 
-  public function userProfile($mobile) 
+  public function userProfile($mobile)
   {
     $response = $this->client->get('?phone_number=' . $mobile . '&include_messages=true');
 
     $xml = $response->xml();
 
     $json = json_encode($xml);
- 
+
     $array = json_decode($json, TRUE);
 
-    if (!empty($array['error'])){
-      return array();
-    } else {
-      return $array['profile'];
+    if (empty($array['error'])){
+      return array_filter($array['profile']);
     }
   }
 
   public function userMessages($mobile)
   {
     $profile = $this->userProfile($mobile);
-
     if (!empty($profile)) {
       return $profile['messages']['message'];
     } else {
--- a/app/Aurora/Services/MobileCommons.php
+++ b/app/Aurora/Services/MobileCommons.php
@@ -43,7 +43,7 @@ class MobileCommonsAPI {
   {
     $profile = $this->userProfile($mobile);
     if (!empty($profile)) {
-      return $profile['messages']['message'];
+      return array_filter($profile['messages']['message']);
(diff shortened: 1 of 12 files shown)


app/Aurora/Services/MobileCommons.php
```diff
@@ -24,29 +24,26 @@ public function __construct()
...
- 
+
     $array = json_decode($json, TRUE);
 
-    if (!empty($array['error'])){
-      return array();
-    } else {
-      return $array['profile'];
+    if (empty($array['error'])){
+      return array_filter($array['profile']);
```
axton21: Using array_filter to filter out empty array from MobileCommons API. Part of PR #88 mobile-commons.blade.php

app/controllers/UsersController.php
```diff
@@ -148,11 +137,11 @@ public function adminCreate($user_id)
 
   public function mobileCommonsMessages($id)
   {
-    $user = $this->northstar->getUser('_id', $id);
+    $northstarUser = new NorthstarUser($id, new NorthstarAPI, new DrupalAPI, new MobileCommonsAPI);
```
angaither: php vars should be `snake_case`

app/controllers/UsersController.php
```diff
@@ -61,28 +63,15 @@ public function store()
...
-          }
-        }
-      }
-    }
-    $mc_profile = $this->mobileCommons->userProfile($user['mobile']);
-
-    $mc_messages = $this->mobileCommons->userMessages($user['mobile']);
-    
-    return View::make('users.show')->with(compact('user', 'aurora_user', 'campaigns', 'reportbacks', 'mc_messages', 'mc_profile'));
+    $northstarUser = new NorthstarUser($id, new NorthstarAPI, new DrupalAPI, new MobileCommonsAPI);
```
DFurnes: Since this is a plain class and not something that Laravel creates for you, it seems that Laravel's IoC container won't automatically resolve dependencies (which is why you have to manually create a `new NorthstarAPI` etc in here).

That gets unwieldy pretty quickly though (imagine if `NorthstarAPI` had injected dependencies of it's own), so I think I'd recommend switching this up to call the IoC container from within `NorthstarUser` to create your dependencies:

```php

use App;

class NorthstarUser {

  public function __construct($id) {
    $this->northstar = App::make('Aurora\Services\Northstar\NorthstarAPI');
    $this->drupal = App::make('Aurora\Services\Drupal\DrupalAPI');
    // ...
  } 
```

axton21:
@DFurnes @angaither

DFurnes:
Looks good to me! :+1:

PR #93, opened 2015-07-23
changed hel helper functions from snake case to camel case

paulhan221:
@DFurnes @angaither 

very small change

### DIFF
--- a/app/Aurora/helpers.php
+++ b/app/Aurora/helpers.php
@@ -30,7 +30,8 @@ function messageStateClass($message)
 }
 
 //https://github.com/giggsey/libphonenumber-for-php
-function sanitizePhoneNumber($number, $countryName = 'US'){
+function sanitizePhoneNumber($number, $countryName = 'US')
+{
 	$phoneUtil = \libphonenumber\PhoneNumberUtil::getInstance();
 	try {
 		$formattedNumber = $phoneUtil->parse($number, $countryName);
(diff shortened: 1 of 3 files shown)


app/Aurora/helpers.php
```diff
@@ -9,7 +9,7 @@ function autoOpenModal(){
 	</script>';
 }
 
-function time_formatter($date)
+function timeFormatter($date)
```
DFurnes: Helper methods should use `snake_case` so this was good before!

app/Aurora/helpers.php
```diff
@@ -18,7 +18,7 @@ function time_formatter($date)
   return date_format($date, 'Y-m-d h:ia');
 }
 
-function message_state_class($message)
+function messageStateClass($message)
```
DFurnes: Same! (The reason being that since helper methods are added to the global scope, it's nice to visually differentiate them from class methods.)

app/Aurora/helpers.php
```diff
@@ -30,7 +30,8 @@ function message_state_class($message)
 }
 
 //https://github.com/giggsey/libphonenumber-for-php
-function sanitizePhoneNumber($number, $countryName = 'US'){
+function sanitizePhoneNumber($number, $countryName = 'US')
```
DFurnes: :no_bicycles:

DFurnes:
I started by commenting on each line and then realized that was silly. :dancer:

PR #95, opened 2015-07-23
changed helper functions from camelCase to snake_case

paulhan221:
@DFurnes @angaither 

I think I got it right this time...


DFurnes:
:+1: :ship:

PR #96, opened 2015-07-24
fixed admin remove route

paulhan221:
@DFurnes  @angaither 

Very minor change, the routes were conflicting so I had to change the route. I thought I had done this in the Admin-fix pull request but I must have undone it before I pushed.

### DIFF
--- a/app/controllers/UsersController.php
+++ b/app/controllers/UsersController.php
@@ -117,7 +117,8 @@ class UsersController extends \BaseController {
    */
   public function destroy($id)
   {
-    //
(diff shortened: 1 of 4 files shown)


angaither:
it might be worth turning this into a resource so it has the post and creates set up already

DFurnes:
Yeah, this should definitely be restful.

paulhan221:
@DFurnes @angaither

DFurnes:
Seems good to me. :+1:

PR #97, opened 2015-07-25
Pagination with numbers

axton21:
@DFurnes @angaither 
Pagination with numbers. Looks much better than before and move it to the bottom of the table.

### DIFF
--- a/app/views/users/partials/waypoints.blade.php
+++ b/app/views/users/partials/waypoints.blade.php
@@ -1,27 +1,27 @@
 <ul class = "waypoints pagination-buttons" id="pagination-box">
   @if ($data['current_page'] > 1)
     <?php $prev = $data['current_page'] - 1 ?>
-    <div class="">
-
-    </div>
-    <li> {{ link_to_route('users.index', 'First', array('page=1'))}} </li>
-    <li> {{ link_to_route('users.index', 'Previous', array('page=' . $prev))}} </li>
+    <li> {{ link_to_route('users.index', 'First', array('page=1')) }} </li>
+    <li> {{ link_to_route('users.index', 'Previous', array('page=' . $prev)) }} </li>
   @endif
-  <?php $lastLeft = ($data['current_page'] - 3 >= 1) ? $data['current_page'] - 3 : 1 ?>
-  <?php $lastRight = ($data['current_page'] + 3 <= $data['last_page']) ? $data['current_page'] + 3 : $data['last_page'] ?>
+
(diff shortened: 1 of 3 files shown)


axton21:
screen shot 2015-07-25 at 01 47 58
screen shot 2015-07-25 at 01 48 45
screen shot 2015-07-25 at 01 48 24

app/views/users/partials/waypoints.blade.php
```diff
@@ -1,15 +1,26 @@
...
+    <?php $prev = $data['current_page'] - 1 ?>
+    <li> {{ link_to_route('users.index', 'First', array('page=1')) }} </li>
+    <li> {{ link_to_route('users.index', 'Previous', array('page=' . $prev)) }} </li>
+  @endif
 
- @if ($data['current_page'] > 1)
-   <?php $prev =  $data['current_page'] - 1 ?>
-   <li> {{ link_to_route('users.index', 'Prev', array('page=' . $prev))}} </li>
- @endif
+  <?php $firstLeft = ($data['current_page'] - 5 >= 1) ? $data['current_page'] - 5 : 1 ?>
```
angaither: snake case these vars `first_page`

app/views/users/partials/waypoints.blade.php
```diff
@@ -1,15 +1,26 @@
...
+  <?php $lastRight = ($data['current_page'] + 4 <= $data['last_page']) ? $data['current_page'] + 4 : $data['last_page'] ?>
+  <?php $lastRight = $lastRight < 10 && $data['last_page'] > 10 ? 10 : $lastRight ?>
+  <?php $firstLeft = $firstLeft > 10 && $data['last_page'] > 10 ? $lastRight - 9 : $firstLeft ?>
```
angaither: also, can you add comments about what this logic is doing

axton21:
@angaither @DFurnes 
so overall with this pagination,  the index page will always has ten numbers.

angaither:
:+1:

DFurnes:
For a future refactoring, this would be a great use case for Laravel's [built-in paginator](http://laravel.com/docs/4.2/pagination).

DFurnes:
(:+1: to merge though! Looks great!)

axton21:
@DFurnes does that works for calling API though?

DFurnes:
I think so! You should be able to make a custom Paginator instance (to use Paginator niceness with the API response) & a [custom presenter](http://laravel.com/docs/4.2/pagination#custom-presenters) to format it using Neue's markup. Definitely not required for this to merge though – just a suggestion for future refactoring. :smile:

axton21:
@DFurnes Sure do, thanks!

Issue #28, opened 2015-06-25
User merge


angaither:
This could be a beast. But the ability to surface duplicate users and merge them together

DFurnes:
We should revisit this functionality at some point – Northstar is no longer creating duplicate users, but we may still have situations where we want to explicitly merge two accounts (for example, when a user has one account with just an email & another with just a mobile number).

One "gotcha" that we'll have to think about is how we combine references to merged IDs in client apps...

angaither:
the idea of this was to deal with users who had two accounts, not the duplicate user problem

DFurnes:
> the idea of this was to deal with users who had two accounts, not the duplicate user problem

Yeah, I remember talking about that. I'm not sure that's how the feature we had built worked though... it looks like it was only allowing you to merge [when duplicate records were found in searches](https://github.com/DoSomething/aurora/pull/98#issuecomment-126098843).

PR #98, opened 2015-07-29
User search

paulhan221:
@DFurnes @angaither 

References #28


paulhan221:
image
displaying search results

paulhan221:
image
It then scrolls down to an edit form with the user attributes merged and the selected user overwriting any attribute that might be different.

paulhan221:
image

Lastly, upon clicking "Merge Users" a confirmation box opens up. The form submits a PUT request to update the selected user, and an ajax call is made to the UsersController to delete the other users that weren't selected. After debating over which was the best and most logical way to go about to solve this problem, Anton and I feel this makes the most sense. Would love to hear thoughts/criticisms!

app/controllers/UsersController.php
```diff
@@ -123,14 +125,16 @@ public function destroy($id)
 
   public function search()
   {
-    $search = Input::get('search_by');
+    $search = filter_var(Input::get('search_by'), FILTER_SANITIZE_STRING);
```
angaither: shouldn't laravel handle this filtering?

app/views/search/merge-and-delete-form.blade.php
```diff
@@ -0,0 +1,62 @@
+<div class ="container -padded">
+  <div class="wrapper">
+    <h1 class="heading -hero">Please verify user info</h1>
+    <div class="container__block -narrow">
+      {{ Form::model($user, [ 'method' => 'PATCH', 'route' => ['users.update', $user['_id']]]) }}
+      <div class="form-item -padded">
+        {{ Form::label('_id', 'Mongo Id', ['class' => 'field-label']) }}
+        {{ Form::text('_id', NULL, ['class' => 'text-field', 'disabled' => 'true']) }}
```
angaither: all of this stuff is the same on most of the user forms, can we break it out into a partial?

app/Aurora/Services/Northstar.php
```diff
@@ -82,4 +82,24 @@ public function createNewApiKey($input)
       ]);
     return $response->json();
   }
+  // used in the search action of the users controller for the sake of duplicate users
+  public function getUsers($type, $id)
```
angaither: you can probably just update the getUser function to handle this instead of creating a second one

app/routes.php
```diff
@@ -23,8 +23,19 @@
 
 # Users
 Route::resource('users', 'UsersController');
+
+# Delete Northstar User
+Route::delete('northstar-user-delete/{user}', ['as' => 'northstar.delete', 'uses' => 'UsersController@deleteNorthstarUser']);
```
angaither: you shouldn't need to specifically create this, it should automatically be handled with the resource above
`Route::resource('users', 'UsersController');`

app/Aurora/Services/Northstar.php
```diff
@@ -82,4 +82,24 @@ public function createNewApiKey($input)
...
+  {
+    $response = $this->client->get('users' . '/' .  $type  . '/' . $id);
+    $northstar_users = $response->json()['data'];
+    // sort users by "updated_at" attribute
```
DFurnes: This should be happening on Northstar (as a parameter on the API query). Fine to do like this for now, but add an `@TODO`(and an associated Northstar issue) to clean this up.

app/Aurora/helpers.php
```diff
@@ -4,12 +4,22 @@ function auto_open_modal()
...
+		});
 	</script>';
 }
 
+function add_class_to_first_result()
+{
+	return
+	"<script>
+		$(document).ready(function () {
+			$('.results').first().addClass('first-result').prepend('<h3>This appears to be the most recent user!</h3>')
```
DFurnes: This can probably be done in your server rendered templates.

app/Aurora/helpers.php
```diff
@@ -42,3 +52,61 @@ function sanitize_phone_number($number, $countryName = 'US')
 		return $number;
 	}
 }
+
+function ajax_edit_merge_form()
+{
+	return 
+	"<script>
+	$(document).ready(function(){
+		$('.merge').click(function(){
```
DFurnes: It's a good practice to add a `.js-` prefix to selectors that you're using for event handlers. That way you don't have to worry about breaking scripts if you change an appearance class and vice versa.

app/Aurora/helpers.php
```diff
@@ -42,3 +52,61 @@ function sanitize_phone_number($number, $countryName = 'US')
...
+				$('html, body').animate({
+					scrollTop: $('#merge-form').offset().top
+				}, 800);
+			});
+		});
+	});
+</script>";
+}
+
+function ajax_confirm_and_delete()
```
DFurnes: It'd be nice if the special "delete" logic & the "confirmation" handler were separated, so you could re-use this in multiple places. Take a look at [`confirm.js`](https://github.com/DoSomething/voting-app/blob/dev/resources/assets/js/utilities/confirm.js) from the Voting App for some inspiration.

app/controllers/UsersController.php
```diff
@@ -2,6 +2,7 @@
 
 use Aurora\NorthstarUser;
 use Aurora\Services\Northstar\NorthstarAPI;
+use Illuminate\Support\Facades\Input;
```
DFurnes: You can just use `use Input;` here. Laravel makes Facades accessible in the global scope.

app/controllers/UsersController.php
```diff
@@ -105,7 +106,8 @@ public function update($id)
   {
     $input = Input::except('_token', '_id', 'drupal_uid');
     $user = $this->northstar->updateUser($id, $input);
-    return Redirect::back()->with('flash_message', ['class' => 'messages', 'text' => 'Sweet, look at you updating that user.']);
+    // return Redirect::back()->with('flash_message', ['class' => 'messages', 'text' => 'Sweet, look at you updating that user.']);
```
DFurnes: Remove this commented line! :)

app/controllers/UsersController.php
DFurnes: +1. Laravel will handle escaping data at the point where the problem could occur (e.g. database or HTML), since escaping rules are different for different cases.

app/Aurora/Services/Northstar.php
paulhan221: image
The getUser function only returns the first result. For search results, we need to display all of them. I'm not sure how we would use the getUser function to display all results without affecting the show page where only one result is needed. =/

app/routes.php
paulhan221: @angaither 
The delete resource for users is being used to delete admins, this route is specifically to delete users in the Northstar db

angaither: ah, got it @paulhan221 hm. 
Perhaps as a future improvement we should break out all Northstar things into it's own NorthstarUserController. But that can come later.

paulhan221:
@angaither @DFurnes

DFurnes:
:ship:!

PR #99, opened 2015-08-03
Zendesk

paulhan221:
References #9 
@DFurnes @angaither


app/Aurora/Services/Zendesk.php
```diff
@@ -0,0 +1,41 @@
...
+
+  public function __construct()
+  {
+    $base_url = \Config::get('services.zendesk.url');
+
+    $username = \Config::get('services.zendesk.username');
+
+    $password = \Config::get('services.zendesk.password');
+
+    $client = new \GuzzleHttp\Client([
```
DFurnes: You can just use `$client = new Client` here, since `GuzzleHttp\Client` is specified in the `use` statement at the top of this file.

app/Aurora/Services/Zendesk.php
```diff
@@ -0,0 +1,41 @@
...
+      'base_url' => $base_url,
+      'defaults' => array(
+      'auth' => [$username, $password, 'Basic']
```
DFurnes: Indentation looks a bit wonky on this line.

DFurnes:
Two little comments, otherwise :ship: :ship: :ship:!

PR #100, opened 2015-08-05
Highlighting on merge user form-fields

antonhalim:
# What's this PR do?
* Adding higlights to form-fields if there is difference in value from duplicate users field in the merge form

# Any involved tickets?
* #28 Duplicate users functionality

___
screen shot 2015-08-05 at 14 49 47


app/Aurora/helpers.php
```diff
@@ -34,6 +34,25 @@ function sanitize_phone_number($number, $countryName = 'US')
...
+  foreach($keep_user as $k => $v){
+    if(empty($delete_user[$k]) && !in_array($k, $array)){
+      array_push($array, $k);
+    }elseif(!empty($delete_user[$k]) && $keep_user[$k] != $delete_user[$k]){
+      array_push($array, $k);
+    }
+  }
+  return $array;
+}
+
```
axton21: @angaither @DFurnes This function is returning an array of all attributes which is different from the account that the user trying to merge, ex: name, mobile, add_city, etc.
It probably needs further refactoring, been trying to use some php function like array_diff, array_diff_uassoc, and array_diff_assoc didnt seem help much to get what we want.

app/Aurora/helpers.php
```diff
@@ -34,6 +34,25 @@ function sanitize_phone_number($number, $countryName = 'US')
...
+  }
+  foreach($keep_user as $k => $v){
+    if(empty($delete_user[$k]) && !in_array($k, $array)){
+      array_push($array, $k);
+    }elseif(!empty($delete_user[$k]) && $keep_user[$k] != $delete_user[$k]){
+      array_push($array, $k);
+    }
+  }
+  return $array;
+}
```
angaither: style, needs spaces after foreach, if, and parens

app/views/search/merge-and-delete-form.blade.php
```diff
@@ -6,6 +6,10 @@
   </div>
 </div>
 
-<script>	
+<script>
+@forelse($different_tags as $tag)
+  $('input[name={{ $tag }}]').addClass("has-error")
```
DFurnes: This should be done in your blade form template, rather than as a client-side script.

DFurnes:
I'd suggest using a different styling for noting conflicts, since the red border makes me think the field has an error (that's what we use that style for elsewhere). Perhaps a yellow border?

axton21:
Okay, I ll fix the styling, but this PR is not up-to-date the most recent commits and will be much better if it's on hold untill #105 is shipped. @DFurnes

axton21:
screen shot 2015-08-20 at 16 20 23

axton21:
@DFurnes @angaither 
Yellow def looks better.

PR #105, opened 2015-08-17
Fix merge

paulhan221:
# What's this PR do?
*  While we were building out the advanced-search functionality, we forgot to take into account the ability to merge duplicate users. This PR fixes that.
* We also added more fields to the advanced-search fields (First name, last name, email, drupal id)

# What are the relevant tickets?
* References #28 

@angaither @DFurnes

### DIFF
--- a/app/Aurora/helpers.php
+++ b/app/Aurora/helpers.php
@@ -118,3 +118,23 @@ function param_builder($input)
   }
   return $query;
 }
+
+function check_if_email_or_mobile($query)
+{
+	if (key($query) == 'email' || key($query) == 'mobile') {
+		return true;
+	} else {
+		return false;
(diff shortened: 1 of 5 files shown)


app/Aurora/helpers.php
```diff
@@ -118,3 +118,23 @@ function param_builder($input)
...
+	}
+}
+
+function duplicate_user_check($array)
+{
+  foreach ($array as $element) {
+    if ($element['email'] == next($array)['email']) {
+    	return true;
+    } else if ($element['mobile'] == next($array)['mobile']) {
+    	return true;
```
angaither: true and false should be caps

angaither:
other than that :+1:

Issue #104, opened 2015-08-14
Opt out user from communications


angaither:
this button can go in the danger zone section
it will make calls to 
- [ ] mobile commons
- [ ] mailchimp
- [ ] drupal?

to remove the user from communications

PR #106, opened 2015-08-18
Mailchimp

paulhan221:
References #104 

This pull request integrates the MailChimp API and implements an "unsubscribe" functionality for MailChimp only.

We are thinking we should make unsubscribe buttons individually for now, and when the drupal endpoint for unsubscribing a user is made we could combine them all into one button?

@DFurnes  @angaither


app/Aurora/NorthstarUser.php
```diff
@@ -141,4 +142,33 @@ public function unassignedRoles($user_roles) {
     return $unassigned_roles;
   }
 
+
+/**
+   * Used in UsersController->show()
```
DFurnes: Looks like indentation got a little borked here! :eyeglasses:

app/routes.php
```diff
@@ -56,3 +56,6 @@
 
 # Unauthorized Page
 Route::get('/unauthorized', 'SessionsController@unauthorized');
+
+# Unsubscribe to MailChimp
+Route::post('unsubscribe-mailchimp', ['as' => 'users.unsubscribe-mailchimp', 'uses' => 'UsersController@unsubscribeFromMailChimp']);
```
DFurnes: Let's consider if we can make this more restful... perhaps something like `DELETE /user/{id}/mailchimp`. Not necessary for merge, but worth thinking about.

app/Aurora/Services/MailChimp.php
```diff
@@ -0,0 +1,43 @@
...
+  {
+    // for "test" users only
+    $this->testID = \Config::get('services.mailchimp.test_id');
+    // international members
+    $this->internationalID = \Config::get('services.mailchimp.international_id');
+    // regular dosomething members
+    $this->domesticID = \Config::get('services.mailchimp.domestic_id');
+    // 26+ club members
+    $this->dinosaurID = \Config::get('services.mailchimp.dinosaur_id');
+    $this->client = new \Drewm\MailChimp(\Config::get('services.mailchimp.apikey'));
```
DFurnes: This is a bit awkward. I'd suggest importing the `MailChimp` API class at the top of the file using a `use` statement. It also might be a good idea to add a `config()` helper method to the app (which is built-in in Laravel 5) as a shortcut for `\Config::get` since it's being used so often.

paulhan221:
made it more restful like you recommended
@DFurnes

angaither:
:+1:

PR #112, opened 2016-01-13
Update to Laravel 5.1.

DFurnes:
#### Changes
Updates Aurora to Laravel 5.1, so it's like our other apps.

For review: @angaither 

img_0309-edit


.env.example
```diff
@@ -0,0 +1,23 @@
...
+DB_USERNAME=homestead
+DB_PASSWORD=secret
+
+CACHE_DRIVER=file
+SESSION_DRIVER=file
+QUEUE_DRIVER=sync
+
+NORTHSTAR_URL=http://northstar.dev:8000
+NORTHSTAR_APP_ID=456
+NORTHSTAR_API_KEY=abc4324
```
weerd: should this be in here?

DFurnes: It's the key we use in our [database seeder](https://github.com/DoSomething/northstar/blob/c9ea278ec3b8509a934be94595caacc7c7840318/database/seeds/ApiKeyTableSeeder.php#L17-L21), so I think it's fine.

angaither: sure! they aren't the real ones!

weerd: :+1:

angaither:
what a day! :100:

PR #114, opened 2016-01-14
Api Keys: allow admins to edit scope and delete

### LINKED ISSUE #23: Delete api keys
Create a `delete` api key functionality in the admin ui

DFurnes:
#### Changes
Wheeeee. Admin users can now set scopes on API keys, modify existing keys, or delete keys altogether without having to muck around in the command line. Phew!

Closes #23.

---
For review: @angaither


app/Http/Controllers/KeyController.php
```diff
@@ -2,48 +2,119 @@
...
 
     /**
      * Display a listing of the resource.
+     * GET /keys
      *
      * @return Response
      */
     public function index()
     {
-        try {
```
angaither: you don't like trying things?

DFurnes: do or do not

DFurnes: More seriously, it just seemed to make things read a bit easier. I think it'd be worth adding better error handling when we get a chance, but I think the only way any of these would fail is if Aurora's API key wasn't set (which breaks other things too).

angaither:
:+1:

PR #115, opened 2016-01-15
Use Laravel authentication controller for auth.

DFurnes:
#### Changes
Uses Laravel’s authentication controller & a custom Authentication driver to handle auth. This removes the need for a lot of boilerplate code, and is a nice proof-of-concept for Northstar as external auth provider.

I also re-indented a bunch of stuff in the `UsersController`, so ignore that messiness in the diff. :grimacing: 

#### Where should the reviewer begin?
I'd recommend looking at [NorthstarUserProvider](https://github.com/DoSomething/aurora/blob/auth/app/Auth/NorthstarUserProvider.php) and the new, slimmer [AuthController](https://github.com/DoSomething/aurora/blob/auth/app/Http/Controllers/Auth/AuthController.php) since that's where most of the action happens. Laravel's [authentication documentation](https://laravel.com/docs/5.1/authentication) is also worth a quick read.

---

For review: @angaither


angaither:
:+1:

PR #117, opened 2016-01-20
API Response model, and other things

DFurnes:
#### Changes
This PR adds a lightweight model class for API responses, so we can treat users delivered as Northstar API responses as if they were a local Eloquent item/collection (with mutators & methods). I _literally_ copied attribute-related methods from the Eloquent model class for this, but this could probably be cleaned up some more in the future and packaged as part of a PHP API. (0a83ff6)

It also removes the code for third-party integrations (MailChimp, Mobile Commons, etc.) until we can revisit them later on. (fc0800f) And a slew of other smaller tweaks/cleanup. (426dc36, fab07ca, 2493bc0, 86f277d).

#### How should this be reviewed?
I'd recommend taking a look commit-by-commit.

---

For review: @angaither


angaither:
:+1:

PR #118, opened 2016-01-20
Pagination

DFurnes:
#### Changes
Fixed a bug in pagination that was causing errors on pages past 5… then redid pagination to use Laravel’s built-in pagination and a custom presenter for Forge markup. I also added an API response collection class for simplifying dealing with “collections” returned from Northstar.

—
For review: @angaither


app/Http/Presenters/ForgePaginationPresenter.php
```diff
@@ -0,0 +1,135 @@
...
+
+    /**
+     * Get HTML wrapper for active text.
+     *
+     * @param  string  $text
+     * @return string
+     */
+    protected function getActivePageWrapper($text)
+    {
+        return '<li class="is-active"><a href="#">'.$text.'</a></li>';
```
angaither: this doesn't look like :otwell:

angaither:
:+1:

app/Http/Presenters/ForgePaginationPresenter.php
DFurnes: Surprising [but true](https://github.com/laravel/framework/blob/e804393253a36d32161c95a7b37639cdb6009b75/src/Illuminate/Pagination/BootstrapThreePresenter.php#L103)!

angaither: well, actually. 
image

DFurnes: GOD DAMN IT, GRAHAM.

weerd: Yaaasss!

PR #119, opened 2016-01-21
Roles clean-up.

DFurnes:
#### Changes
Cleans up implementation of Aurora user roles.

Since users can only have one role at any given time, using a pivot table seemed to complicate things more than necessary. This should fix some bugs we were having with updating user roles on production.

I also renamed the `_id` property on Aurora users to `northstar_id` to be more clear about its purpose.

---
For review: @angaither @weerd


weerd:
:sheep: it!

PR #120, opened 2016-01-21
Cleanup old roles tables.

DFurnes:
#### Changes
Cleanup from changes made in #119. I left this as a separate PR so that we could more easily revert if the changes made in #119 cause any problems on deploy and need to be reverted.

---
For review: @angaither


DFurnes:
And everything looks good on QA & production post-deploy, so this should be safe to merge! :soon:

angaither:
woah! :bomb: :ship:

PR #121, opened 2016-01-21
Grant permission to multiple roles.

DFurnes:
#### Changes
Allow a route/controller to be limited to any of an array of roles (e.g. users can be viewed by admins _or_ staff _or_ interns). :hand: 

---
For review: @angaither or @weerd

### DIFF
--- a/app/Http/Middleware/CheckRole.php
+++ b/app/Http/Middleware/CheckRole.php
@@ -29,11 +29,11 @@ class CheckRole
      *
      * @param  \Illuminate\Http\Request $request
      * @param  \Closure $next
-     * @param $roles
      * @return mixed
      */
     public function handle($request, Closure $next)
     {
+        // e.g. $this->middleware('role:admin,intern') => ['admin', 'intern']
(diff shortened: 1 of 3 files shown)


app/Http/Middleware/CheckRole.php
```diff
@@ -27,14 +27,17 @@ public function __construct(Guard $auth)
     /**
      * Handle an incoming request.
      *
-     * @param  \Illuminate\Http\Request  $request
-     * @param  \Closure  $next
+     * @param  \Illuminate\Http\Request $request
+     * @param  \Closure $next
+     * @param $roles
```
weerd: Remove the `@param` no longer a part of passed params...

weerd:
Cool! Just the one note, otherwise :boat: :+1:

app/Http/Middleware/CheckRole.php
DFurnes: Ah, fixed and added a comment to the line below.

PR #124, opened 2016-01-25
Route model binding

DFurnes:
#### Changes
I think this pretty much wraps up the "tidying" PRs for Aurora.

This updates all controllers to use [route model binding](https://laravel.com/docs/5.1/routing#route-model-binding) to inject relevant models into the controllers (rather than having to re-write the same fetch logic in each method). It also adds a `NorthstarKey` class to wrap API key responses, and a spiffy new 404 page.

---
For review: @angaither


angaither:
:+1:

PR #126, opened 2016-01-26
Update edit route to expect user, rather than just ID.

DFurnes:
#### Changes
Fixes an error on editing a user, since it is provided a full `NorthstarUser` object by the route model binding (added in #124) but the `edit`, `update`, and `destroy` methods still expected only a string ID.

---

For review: @angaither


angaither:
:+1:

PR #127, opened 2016-02-02
Add RestAPIClient to clean up API classes.

DFurnes:
##### REST API client cleanup
Adds a small helper class to clean up calls to RESTful APIs so we don’t have to have so much boilerplate. Also updates documentation!

##### Properly handles validation errors
Aurora would previously explode :bomb: if it received a `422` response from Northstar. We now explicitly catch any Guzzle client exceptions and, if it's a validation error, format and display them.

screen shot 2016-02-02 at 1 28 07 pm

##### Add a nice 500 page
This also adds a nice 500 page (which will be used for any `HttpExceptions`, such as any uncaught exception returned from the Northstar API). This does _not_ replace fatal errors (like calling an nonexistent method), which are still sent to the Symphony "whoops" reporter.

---
For review: @angaither @weerd

### DIFF
--- a/app/APIResponseCollection.php
+++ b/app/APIResponseCollection.php
@@ -14,20 +14,28 @@ class APIResponseCollection extends Collection
      */
     protected $paginator;
 
-    public function __construct($response)
+    /**
+     * APIResponseCollection constructor.
+     * @param array $response - Array of raw API responses
+     * @param string $class - Class to create for contents
+     */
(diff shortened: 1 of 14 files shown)


weerd:
Lots o' stuff! @_@ But seems cool :sheep: it!

DFurnes:
:dancer:

PR #128, opened 2016-02-04
Update for northstar auth

DFurnes:
#### Changes
Includes updates for Northstar authentication changes in DoSomething/northstar#268.

Other notable changes:

* Removes casting from the API model class for integers, floats, strings, and booleans. I had copied this over when I extracted this stuff from Eloquent's model class, but it doesn't really make any sense to have since JSON supports all the "cartable" data types natively. Also cleans up date casting.
* Move "formatted" number helper into User model (rather than as a global helper method).
* Return `null` when an endpoint 404s. This saves us from having more try/catch blocks than we need to, and seems like a reasonable way to handle this kind of thing.
* Make `Northstar` API class a singleton and inject config values through the service container (which will help us eventually break this class out into a standalone Northstar API package!)
* Adds some missing docblocks and stuff.

---
For review: @angaither / @weerd

### DIFF
--- a/resources/views/users/show.blade.php
+++ b/resources/views/users/show.blade.php
@@ -18,11 +18,13 @@
                 <dt>First Name:</dt><dd>{{ $user->first_name or '&mdash;' }}</dd>
                 <dt>Last Name:</dt><dd>{{ $user->last_name or '&mdash;' }}</dd>
                 <dt>Email:</dt><dd>{{ $user->email or '&mdash;' }}</dd>
-                <dt>Mobile:</dt><dd>{{ $user->mobile or '&mdash;' }}</dd>
+                <dt>Mobile:</dt><dd>{{ $user->prettyMobile('&mdash;') }}</dd>
                 <dt>Birthdate:</dt><dd>{{ $user->birthdate or '&mdash;' }}</dd>
 
                 @if (isset($user->addr_street1) || isset($user->addr_street2) || isset($user->addr_city) || isset($user->addr_state) || isset($user->addr_zip) )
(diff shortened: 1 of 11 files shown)


resources/views/users/show.blade.php
```diff
@@ -18,11 +18,13 @@
...
                 <dt>Last Name:</dt><dd>{{ $user->last_name or '&mdash;' }}</dd>
                 <dt>Email:</dt><dd>{{ $user->email or '&mdash;' }}</dd>
-                <dt>Mobile:</dt><dd>{{ $user->mobile or '&mdash;' }}</dd>
+                <dt>Mobile:</dt><dd>{{ $user->prettyMobile('&mdash;') }}</dd>
                 <dt>Birthdate:</dt><dd>{{ $user->birthdate or '&mdash;' }}</dd>
 
                 @if (isset($user->addr_street1) || isset($user->addr_street2) || isset($user->addr_city) || isset($user->addr_state) || isset($user->addr_zip) )
                     <dt>Address:</dt><dd>{{ $user->addr_street1 or '' }} {{ $user->addr_street2 or '' }} {{ $user->addr_city or '' }} {{ $user->addr_state or '' }} {{ $user->addr_zip or '' }}</dd>
+                @else
+                    <dt>Address:</dt><dd>&mdash;</dd>
```
angaither: what in the what is a mdash. mad dash

angaither:
kewl :ship:

resources/views/users/show.blade.php
DFurnes: the maddest dash &mdash;

weerd:
:+1:

DFurnes:
:bow:

PR #129, opened 2016-02-22
Use Northstar PHP library for API connection.

DFurnes:
#### Changes
Extracts all the Northstar API classes into a shared library… see DoSomething/northstar-php#2.

---
For review: @angaither or @weerd


weerd:
:+1: coolio!

PR #131, opened 2016-03-14
Fix pagination on search page.

DFurnes:
#### Changes
Fixes an issue paginating the search results page (due to API changes made in Northstar PHP package).

---

For review: @angaither


angaither:
:+1:

PR #136, opened 2016-07-11
Update to use v1.0.0-rc5 of Northstar client.

DFurnes:
#### Changes
Updates to use v1.0.0-rc5 of the Northstar client. See DoSomething/northstar-php#20.

---

For review: @weerd


weerd:
💥 👍

PR #137, opened 2016-07-13
Admin touchups

DFurnes:
#### Changes
Takes advantage of the new public key route and refresh token count added in DoSomething/northstar#389.

---
For review: @weerd @angaither


angaither:
🔑 🌠

PR #138, opened 2016-07-14
Aurora admin touchups

DFurnes:
#### Changes
This PR includes some updates to the Aurora admin experience:

💁 Allows staff to use Aurora by requesting the `role:staff` scope when authenticating.

🐧 Change the "Aurora user" view to "Superusers", which now shows all users  with the admin or staff role _on Northstar_, rather than just the subset in the local Aurora database.

🔏 Adds the Northstar `role` field to the user edit screen, so user roles can be easily changed.

📝 Expose the "title" and "description" fields for OAuth clients. These can be used for annotating client details (since `client_id` can no longer be changed without breaking auth).

💻 Moves "connected accounts" to their own section on the profile, in anticipation of potentially hooking up more third-party services in the future.

---
For review: @weerd @angaither


resources/views/users/show.blade.php
```diff
@@ -80,19 +68,20 @@
...
-                    <dt>Role:</dt> <dd>{{ !empty($auroraUser->role) ? $auroraUser->role : '&mdash;' }}</dd>
-                </div>
-
-                @if(Auth::user()->hasRole('admin') || Auth::user()->hasRole('staff'))
-                    <div class="container__block">
-                        <a class="secondary" href="{{ url('aurora-users/' . $auroraUser->id . '/edit') }}">Update Aurora profile</a>
-                    </div>
+            <div class="container__block -narrow profile-settings">
+                <h3>Connected Accounts</h3>
+                @if(! empty($user->drupal_id))
```
angaither: just a thought (though I'm not sure _how_ we would do this) but it would be cool to see something about compeitions here

angaither:
🚢 !

resources/views/users/show.blade.php
DFurnes: Oh that's a good idea. Can you get a user's Gladiator profile by their Northstar ID?

angaither: yip!

DFurnes: Cool, I think we could just link to that then (and just make sure Gladiator gives an okay 404 if a user has never participated in a competition).

DFurnes: Added #141!

PR #140, opened 2016-07-25
Use shared Webpack config & Modernizr v3.

DFurnes:
### Changes
Updates Aurora to use Forge 6.7, Modernizr v3, and our shared [Webpack config](https://github.com/DoSomething/webpack-config). This helps standardize our build system with our other applications.

Aside from breaking things into partials (rather than the one big CSS file) and removing unused classes and scripts, there are no Aurora style changes in this PR. 🎨 

For review: @DoSomething/front-end

PR #142, opened 2016-07-26
Add the danger zone partial.

DFurnes:
### Changes
I broke this out into a partial but forgot to import it to the main stylesheet in #140. Oy.

welcome to the... danger zone

---
For review: @sbsmith86


sbsmith86:
Whoops! 💯

PR #144, opened 2016-07-27
More fixes

DFurnes:
#### Changes
More fixes and small improvements, including changes related to DoSomething/northstar-php#23.

🔜 Redirects to the correct path after creating a new OAuth client (rather than 404ing).

📟 Adds pagination for the clients index (thanks to using the new `v2/clients` endpoint in Northstar-PHP now).

📄 Adds "title" and "description" fields when creating a new client.

📘 Show the clients' titles on the index page (rather than the potentially less clear Client ID).

☕ Adds the JS helpers from Voting App so that we can have nice clickable "form buttons" and "Did you mean to do this thing?" confirmation alerts.

---
For review: @angaither

### DIFF
--- a/resources/views/clients/create.blade.php
+++ b/resources/views/clients/create.blade.php
@@ -2,19 +2,30 @@
 
 @section('main_content')
 
-    @include('layout.header', ['header' => 'API Keys', 'subtitle' => 'Northstar application access & permissions'])
+    @include('layout.header', ['header' => 'OAuth Clients', 'subtitle' => 'Northstar application access & permissions'])
 
     <div class="container -padded">
         <div class="wrapper">
             <div class="container__block -narrow">
-                <h1>Create Clients</h1>
+                <h1>Create Client</h1>
 
                 @include('layout.errors')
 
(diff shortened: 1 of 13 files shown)


resources/views/clients/create.blade.php
```diff
@@ -2,19 +2,30 @@
...
+                    <div class="form-item -padded">
+                        {!! Form::label('description', 'Description', ['class' => 'field-label']) !!}
+                        {!! Form::textarea('description', null, ['class' => 'text-field', 'placeholder' => 'Explain what this client is used for!']) !!}
+                    </div>
+
+                    <div class="form-item -padded">
                         {!! Form::label('client_id', 'Client ID', ['class' => 'field-label']) !!}
-                        {!! Form::text('client_id', null, ['class' => 'text-field']) !!}
+                        {!! Form::text('client_id', null, ['class' => 'text-field', 'placeholder' => 'client-id']) !!}
+                        <span class="footnote">Careful, this cannot be changed later!</span>
```
angaither: 😱

angaither:
🚢

PR #148, opened 2016-09-29
Northstar-PHP 1.0.0 RC9

DFurnes:
#### Changes
This PR updates Aurora to use the latest Northstar-PHP, and also makes a few little clean-up commits.

---
For review: @weerd

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -38,7 +38,7 @@ class UsersController extends Controller
             'path' => 'users',
         ]);
 
-        return view('users.index')->with(compact('users'));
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -3,7 +3,6 @@
 namespace Aurora\Http\Controllers;
 
 use Aurora\Models\AuroraUser;
-use Aurora\Services\Drupal;
```
weerd: Was this not used anywhere?

DFurnes: Nope! The class doesn't exist either.

Issue #158, opened 2017-09-27
Expose SMS fields


aaronschachter:
It'd be really handy to expose the fields Gambit's updating (https://github.com/DoSomething/northstar/pull/631) to make testing easier (especially for non-devs).

PR #161, opened 2017-10-11
Cleanup!

DFurnes:
This pull request includes general code cleanups - deleting old things we no longer use, adds SMS fields (#158), and fixing an issue where non-admins would see a "Whoops!" if they tried to edit role.


app/Http/helpers.php
```diff
@@ -145,3 +23,17 @@ function csv_to_array($string)
...
     return array_map('trim', explode(',', $string));
 }
+
+/**
+ * Format a string of Markdown into HTML.
+ *
+ * @param $source
+ * @return string
+ */
+function markdown($source)
```
weerd: 🎉

aaronschachter:
Would it make sense to make the SMS Status field editable to admins? It'd be useful for internal testing at least - https://github.com/DoSomething/gambit-conversations/pull/208

DFurnes:
I'm into that!

PR #168, opened 2017-10-20
Fix fatal error on unauthorized page.

DFurnes:
### Changes
You can no longer return a `Illuminate\View\View` directly from middleware in Laravel 5.2+, so this was causing a rather nasty Whoops error. Instead, use the `response()` helper to create an actual `Illuminate\Http\Response`.


mendelB:
whoah, great minds....

PR #170, opened 2017-10-20
Lazy load the user count on the index page.

DFurnes:
### Changes

This pull request adds lazy-loading for the user count on the index page.

screen recording 2017-10-20 at 02 39 pm

📄 Adds `scriptify` helper & uses it to expose user's access token to scripts. 6060360

🤖 Configures Northstar resource server using Gateway. 81f0ed6

#️⃣ Adds a API route for the user count, and a little front-end widget to show it. ce1b6e0

This is _totally_ over-engineered for the use case, but is intended a little proof-of-concept for doing the same thing in Rogue and other applications in the future. Since this endpoint is secured using a Northstar token, it could be used by any other application, in addition to Aurora web requests!

### DIFF
--- a/resources/assets/users/lazy-user-count.js
+++ b/resources/assets/users/lazy-user-count.js
@@ -1,0 +1,21 @@
+import { RestApiClient } from '@dosomething/gateway';
+
+/**
+ * Lazy load user count if it's on the page.
+ */
+async function initialize() {
+  const element = document.getElementById('lazy-user-count');
+  if (! element) {
+    return;
+  }
+
+  const aurora = new RestApiClient(window.location.origin, {
+    headers: { 'Authorization' : `Bearer ${window.AUTH}`}
+  });
+
+  const response = await aurora.get('api/total');
+  element.innerText = response ? response['total'] : '???';
+  element.classList = 'lazy-loading is-loaded';
+}
+
+export default { initialize };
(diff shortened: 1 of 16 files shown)


resources/assets/users/lazy-user-count.js
```diff
@@ -0,0 +1,21 @@
...
+ * Lazy load user count if it's on the page.
+ */
+async function initialize() {
+  const element = document.getElementById('lazy-user-count');
+  if (! element) {
+    return;
+  }
+
+  const aurora = new RestApiClient(window.location.origin, {
+    headers: { 'Authorization' : `Bearer ${window.AUTH}`}
```
itsjoekent: Not against doing this, but I am curious why your [opinion changed on it](https://github.com/DoSomething/phoenix-next/pull/44#discussion_r103316827)?

DFurnes: Because we're able to safely verify the token on the server now (via Northstar's public key w/ [Gateway's packaged server implementation](https://github.com/DoSomething/gateway/releases/tag/v1.7.0)). This was always the end goal, but there were some downsides to the approach you'd taken in that pull request (for example, tokens wouldn't expire).

itsjoekent: oh ok, gotcha! 🌮

PR #171, opened 2017-10-23
Add setup command powered by Environmentalist.

DFurnes:
I made a [fun thing](https://github.com/DFurnes/environmentalist) over the weekend, let's use it here!

screen shot 2017-10-23 at 11 33 56 am


DFurnes:
> (But you make this after I suffer through setting up the boring way?!)

😉

PR #174, opened 2017-10-23
User merge

mendelB:
- Adds new user profile partial to display user profile fields 🏑 
- Adds form to user show page to input dup user id if for merging 🦆 
- Controller to handle previewing and confirming user merge 🎁 
- Actual merging post-confirmation is not yet implemented, pending some work on errors in `Northstar`🎢 
- This is pretty much all Dave but I shall take the credit shamelessly :trollface: 

https://www.pivotaltracker.com/story/show/148973339

### DIFF
--- a/app/Http/Controllers/MergeController.php
+++ b/app/Http/Controllers/MergeController.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Aurora\Http\Controllers;
+
+use Illuminate\Http\Request;
+use DoSomething\Gateway\Resources\NorthstarUser;
+
+class MergeController extends Controller
+{
+    /**
+     * Create a MergeController.
+     */
+    public function __construct()
+    {
+        $this->middleware('auth');
+        $this->middleware('role:admin,staff');
+    }
+
+    /**
+     * Show the form for merging two accounts.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function create(NorthstarUser $user, Request $request)
+    {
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/MergeController.php
```diff
@@ -0,0 +1,47 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function create(NorthstarUser $user, Request $request)
+    {
+        $mergeEndpoint = 'v1/users/'.$user->id.'/merge?pretend=true';
+
+        $response = gateway('northstar')->post($mergeEndpoint, ['id' => $request->query('id')]);
+        $mergedUser = new NorthstarUser($response['data']);
+
+        // - Should we be deleting the merged account?
```
DFurnes: The answer to this is "no"! We don't want to break any references to the Northstar ID, so the "merge" endpoint in Northstar will keep a user account with a "fake" `merged-account-{new_id}@dosomething.invalid` email.

In the future, this is something we probably want to replace with a soft-deleted model.

app/Http/Controllers/MergeController.php
```diff
@@ -0,0 +1,47 @@
...
+        // - merging can be somewhat inelegant unless *no* fields overlap
```
DFurnes: Agreed! Once this is functioning in Aurora, we can start to think about smarter "merge rules" for different fields in the Northstar backend (for example, if a field exists on both profiles but is the same we can obviously merge... and in some cases, we can probably just pick the "newest" non-null field & not worry too much).

PR #193, opened 2018-07-03
Force HTTPS traffic on production.

DFurnes:
This pull request adds a global middleware to Aurora to force HTTPS and redirect traffic to the canonical application URL, e.g. https://dosomething-aurora-dev.herokuapp.com/ → https://admin-dev.dosomething.org.

References DoSomething/devops#388.

### DIFF
--- a/app/Http/Middleware/ForceHttps.php
+++ b/app/Http/Middleware/ForceHttps.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Aurora\Http\Middleware;
+
+use Closure;
+use League\Uri\Schemes\Http;
+use function League\Uri\parse;
+
+class ForceHttps
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @param  \Closure $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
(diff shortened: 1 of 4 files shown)


app/Http/Middleware/ForceHttps.php
```diff
@@ -0,0 +1,40 @@
...
+    public function handle($request, Closure $next)
+    {
+        // If running in a production environment, redirect all traffic to HTTPS
+        // and the correct canonical URL, e.g. never '*.herokuapp.com'!
+        if (config('app.env') == 'production') {
+            $canonicalHost = parse(config('app.url'))['host'];
+            $hasIncorrectHost = $request->header('Host') !== $canonicalHost;
+            $isInsecure = ! $request->secure();
+
+            if ($hasIncorrectHost || $isInsecure) {
```
katiecrane: At first I was like...shouldn't this be if it's not secure?? Then I noticed that it's `$isInsecure` and not `$isSecure`. Words are hard!

DFurnes: Ahhh haha yeah that is kinda weird.

DFurnes: Inlined that conditional in 7b36083... I think it reads a bit easier!

PR #202, opened 2018-09-27
Surface voter registration status in Aurora

katiecrane:
#### What's this PR do?
Surfaces `voter_registration_status` on a user in Aurora.

Looks like this (don't worry, randomly generated test user!):
image

#### How should this be reviewed?
We want this?

#### Relevant Tickets
🤠

### DIFF
--- a/resources/views/users/partials/profile.blade.php
+++ b/resources/views/users/partials/profile.blade.php
@@ -16,3 +16,4 @@
 <dt>SMS Paused:</dt><dd>{{ $user->sms_paused ? '✔' : '✘' }}</dd>
 <dt>Country:</dt><dd>{{ $user->country or '&mdash;' }}</dd>
 <dt>Role:</dt><dd>{{ $user->role or '&mdash;' }}</dd>
+<dt>Voter Registration Status:</dt><dd>{{ $user->voter_registration_status or '&mdash;' }}</dd>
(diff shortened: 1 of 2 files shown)


resources/views/users/partials/profile.blade.php
```diff
@@ -16,3 +16,4 @@
 <dt>SMS Paused:</dt><dd>{{ $user->sms_paused ? '✔' : '✘' }}</dd>
 <dt>Country:</dt><dd>{{ $user->country or '&mdash;' }}</dd>
 <dt>Role:</dt><dd>{{ $user->role or '&mdash;' }}</dd>
+<dt>Voter Registration Status:</dt><dd>{{ $user->voter_registration_status or '&mdash;' }}</dd>
```
chloealee: this is awesome, @katiecrane ! Just curious - what is `&mdash`?

katiecrane:
@DFurnes styleCI is failing because of updates in `Fastly.php` from another PR - is it safe to make the changes to styleci passes?

resources/views/users/partials/profile.blade.php
katiecrane: @chloealee it's an em dash like this —, what we use if we got nothing to show!

DFurnes:
Oh gosh! Yes please! StyleCI was 500ing when I merged that and forgot to re-run later.

resources/views/users/partials/profile.blade.php
chloealee: awesome, thanks for the explanation!

PR #203, opened 2018-10-15
Add commands to import/export from Fastly, and some admin updates.

DFurnes:
This pull request adds `fastly:export` and `fastly:import` commands which allow us to clone dictionaries between Fastly properties (most importantly, from our current production property to the new Terraform-managed one) but also could be used to create backups of this data!

I've also fixed a few bugs with the admin flow:
 - Trying to create a redirect that already exists will now overwrite the existing one rather than 500ing.
 - Redirects can no longer contain query strings (because that's not supported in our new simplified VCL) and are automatically converted to lowercase (since all redirects are case-insensitive now).

### DIFF
--- a/app/Services/Fastly.php
+++ b/app/Services/Fastly.php
@@ -75,15 +75,13 @@ class Fastly extends RestApiClient
             $path = '/'.$path;
         }
 
-        // Create a record in the redirects dictionary.
-        $redirect = $this->post('dictionary/'.$this->redirects.'/item', [
-            'item_key' => $path,
+        // Create or update a record in the redirects dictionary.
+        $redirect = $this->put('dictionary/'.$this->redirects.'/item/'.$path, [
             'item_value' => $target,
         ]);
 
-        // Create a record in the statuses dictionary.
-        $type = $this->post('dictionary/'.$this->types.'/item', [
-            'item_key' => $path,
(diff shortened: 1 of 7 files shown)


app/Services/Fastly.php
```diff
@@ -70,20 +70,18 @@ public function getRedirect($id)
      */
     public function createRedirect($path, $target, $status)
     {
-        // Make sure path begins with a slash.
+        // Make sure path is lower-case & begins with a slash.
         if ($path[0] !== '/') {
-            $path = '/'.$path;
+            $path = strtolower('/'.$path);
```
mendelB: Did you want to lower case *all* paths or just ones that didn't start with a `/`?

DFurnes: All paths! We always [transform paths to lowercase](https://github.com/DoSomething/infrastructure/blob/0f2824edb2dc3ce6158f0ca0c2a1d03ee4e9e4dd/dosomething/fastly-frontend/redirect_recv.vcl#L3) now before checking for them in the redirects dictionary, so any redirects that were created in Aurora with capitalization would be kaput.

DFurnes: And oh man, I see what you mean now! I'll fix that up in a follow-up PR.

PR #207, opened 2019-02-26
Add email subscription status and topics

katiecrane:
### What's Changed

Added `email_subscription_status` and `email_subscription_topics` to user profile. I separated subscriptions out into a separate sections because I thought the list was getting a little long and hard to parse, but as you know I am _not_ a designer so happy to leave everything in one list if folks like that better!

User with no email topics:
image

User with email topics:
image

### DIFF
--- a/resources/views/users/show.blade.php
+++ b/resources/views/users/show.blade.php
@@ -11,8 +11,15 @@
                 @include('layout.errors')
             </div>
             <div class="container__block -half profile-settings">
-                <h3>Profile</h3>
-                @include('users.partials.profile')
+                <div class="container -padded">
+                    <h3>Profile</h3>
+                    @include('users.partials.profile')
+                </div>
+                <div class="container -padded">
+                    <h3>Subscriptions</h3>
+                    @include('users.partials.subscriptions')
+                </div>
+
             </div>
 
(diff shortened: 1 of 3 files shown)


resources/views/users/show.blade.php
```diff
@@ -11,8 +11,15 @@
...
             <div class="container__block -half profile-settings">
-                <h3>Profile</h3>
-                @include('users.partials.profile')
+                <div class="container -padded">
+                    <h3>Profile</h3>
+                    @include('users.partials.profile')
+                </div>
+                <div class="container -padded">
+                    <h3>Subscriptions</h3>
+                    @include('users.partials.subscriptions')
```
DFurnes: Is this partial missing from the PR? 🕵️

katiecrane: agh well I was trying to push it up but GitHub has started asking me for a password each time and sometimes I miss it! it's here now!

DFurnes: Ah, frustrating! You should be able to fix that [by configuring a helper](https://help.github.com/en/articles/caching-your-github-password-in-git) or updating your git remote to use SSH for authentication instead (which is what I do).

PR #211, opened 2019-03-12
Password Reset form

aaronschachter:
Adds a new Danger Zone ™️ form for executing `POST /v/2/resets` requests, allowing admins an easy way to test Activate Account emails that'd otherwise be triggered via Chompy imports or Postman requests. Blocked by https://github.com/DoSomething/gateway/pull/122

Screen Shot 2019-03-12 at 12 03 16 PM

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -146,7 +146,7 @@ class UsersController extends Controller
     {
         $type = $request['type'];
 
-        $response = $this->northstar->sendUserPasswordReset($user->id, $type);
+        $this->northstar->sendUserPasswordReset($user->id, $type);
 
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -134,4 +134,20 @@ public function search(Request $request)
...
+     *
+     * @param NorthstarUser $user
+     * @param Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function sendPasswordReset(NorthstarUser $user, Request $request)
+    {
+        $type = $request['type'];
+
+        $response = $this->northstar->sendUserPasswordReset($user->id, $type);
```
DFurnes: Since `$response` doesn't seem to be used here, let's not save this to variable.

PR #212, opened 2019-03-21
Adds email_subscription_topics to user edit form

aaronschachter:
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/164769906.

### DIFF
--- a/resources/views/users/edit.blade.php
+++ b/resources/views/users/edit.blade.php
@@ -79,14 +79,21 @@
           </div>
           <div class="form-item -padded">
               {!! Form::label('email_subscription_topics', 'Email Subscription Topics', ['class' => 'field-label']) !!}
-                  <div class="select">
-                    {!! Form::select(
-                      'email_subscription_topics', [
-                        'news' => 'news',
-                        'scholarships' => 'scholarships',
-                        'lifestyle' => 'lifestyle',
-                        'community' => 'community',
-                      ], null, ['multiple' => 'multiple', 'name' => 'email_subscription_topics[]']) !!}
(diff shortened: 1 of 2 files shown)


resources/views/users/edit.blade.php
```diff
@@ -77,6 +77,18 @@
                       ], null, ['placeholder' => '--']) !!}
                   </div>
           </div>
+          <div class="form-item -padded">
+              {!! Form::label('email_subscription_topics', 'Email Subscription Topics', ['class' => 'field-label']) !!}
```
DFurnes: I wonder if folks know how to use a multiple-select box. It might be worth adding a little `<p class="footnote">` below explaining that you need to command-click the items. (Or should these be checkboxes?)

aaronschachter: I was struggling with the checkboxes a little bit, but probably doing something silly... this code would save values correctly, but unchecking them all wouldn't clear out the values saved.

```
<div>
  {!! Form::label('news', 'news') !!}
  {!! Form::checkbox('email_subscription_topics[]', 'news') !!}
</div>
<div>
  {!! Form::label('scholarships', 'scholarships') !!}
  {!! Form::checkbox('email_subscription_topics[]', 'scholarships') !!}
</div>
```

DFurnes: Ahh yeah, that's a frustrating gotcha. We ran into the same thing for the OAuth client form in #167... basically, the browser just omits the field entirely when nothing is checked, and so Laravel doesn't try to change it unless you add a little extra logic in the controller.

PR #219, opened 2019-08-02
Add Voter Registration Status field to Profile Edit Form

mendelB:
## What's this PR do?

This PR adds the `voter_registration_status` field to the User's profile edit form.

https://github.com/DoSomething/aurora/pull/219/commits/d52fcbd472779266312e02ae0fa07c43460510b1 Removes a hack to strip the `sms_status` field from the Northstar payload if `null` since we've made that field `nullable`.

## How should this be reviewed?
👀 

## Any background context you want to provide?
This won't actually completely work until we merge https://github.com/DoSomething/northstar/pull/903 to remove some hierarchy validation for arbitrary status updates and make this field `nullable` within Northstar's [Registerer validation](https://github.com/DoSomething/northstar/pull/903/files#diff-78498fe993d8f7dfd5cff7adddcdad3fR69) to allow submitting this field when it's unset or to actively unset this field.

And @katiecrane sorry to belabor this point, but I just want to confirm that you think it's the correct call to have us send along a `null` value for this field instead of restraining it in Aurora so we don't unset anything in Northstar? (One new thought I had is that it might be indirect to allow what is formally a placeholder value to actually update the database?)

## Relevant Tickets

[Pivotal ID #165974120](https://www.pivotaltracker.com/story/show/165974120)

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -74,11 +74,6 @@ class UsersController extends Controller
     {
         $input = $request->except('_token', '_id', 'drupal_uid');
 
-        // @HACK: This field should be `nullable` in Northstar.
-        if (is_null($input['sms_status'])) {
-            unset($input['sms_status']);
-        }
-
         $input['email_subscription_topics'] = ! empty($input['email_subscription_topics']) ? $input['email_subscription_topics'] : [];
 
         $this->northstar->updateUser($user->id, $input);
--- a/resources/views/users/edit.blade.php
+++ b/resources/views/users/edit.blade.php
@@ -78,6 +78,18 @@
                   </div>
           </div>
           <div class="form-item -padded">
+              {!! Form::label('voter_registration_status', 'Voter Registration Status', ['class' => 'field-label']) !!}
+                  <div class="select">
+                      {!! Form::select('voter_registration_status', [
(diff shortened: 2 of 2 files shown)


katiecrane:
> I just want to confirm that you think it's the correct call to have us send along a null value for this field instead of restraining it in Aurora so we don't unset anything in Northstar? (One new thought I had is that it might be indirect to allow what is formally a placeholder value to actually update the database?)

I know we've talked about this a lot already but these sentences are confusing me!

mendelB:
> > I just want to confirm that you think it's the correct call to have us send along a null value for this field instead of restraining it in Aurora so we don't unset anything in Northstar? (One new thought I had is that it might be indirect to allow what is formally a placeholder value to actually update the database?)
> 
> I know we've talked about this a lot already but these sentences are confusing me!

I started making a little more sense in Slack in [this thread](https://dosomething.slack.com/archives/C02BBRN5A/p1564769172001100)

We determined that we can keep this field `nullable`, and while doing so had the chance to update the `sms_status` field to `nullable` as well in https://github.com/DoSomething/northstar/pull/903 and remove an old hack in https://github.com/DoSomething/aurora/pull/219/commits/d52fcbd472779266312e02ae0fa07c43460510b1

PR #221, opened 2019-09-19
Edit refer-friends feature flag

aaronschachter:
This PR adds support to edit the `refer-friends` feature flag value introduced in https://github.com/DoSomething/northstar/pull/915.

#220 did not let admins remove the `badges` feature flag by disabling the input, but it's pretty tricky to keep this logic for two checkboxes: If a user has the badges flag, the input is disabled. We want to give them the refer-friends feature flag, save, but then the user no longer has the badges flag because the input was disabled. I decided to go for a simpler approach by just making both flags editable. cc @ngjo

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -76,8 +76,11 @@ class UsersController extends Controller
 
         $input['email_subscription_topics'] = ! empty($input['email_subscription_topics']) ? $input['email_subscription_topics'] : [];
 
-        if (array_key_exists('feature_flags', $input) && in_array('badges', $input['feature_flags'])) {
-            $input['feature_flags'] = ['badges' => true];
+        if (array_key_exists('feature_flags', $input)) {
+            $input['feature_flags'] = [
+                'badges' => in_array('badges', $input['feature_flags']),
+                'refer-friends' => in_array('refer-friends', $input['feature_flags']),
+            ];
         } else {
             unset($input['feature_flags']);
         }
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -76,8 +76,11 @@ public function update(NorthstarUser $user, Request $request)
 
         $input['email_subscription_topics'] = ! empty($input['email_subscription_topics']) ? $input['email_subscription_topics'] : [];
 
-        if (array_key_exists('feature_flags', $input) && in_array('badges', $input['feature_flags'])) {
-            $input['feature_flags'] = ['badges' => true];
+        if (array_key_exists('feature_flags', $input)) {
+            $input['feature_flags'] = [
+                'badges' => in_array('badges', $input['feature_flags']),
+                'refer-friends' => in_array('refer-friends', $input['feature_flags']),
```
mendelB: _not to hold up this merge, but perhaps an idea for a post-launch PR_:

Re disabling the input to prevent admins unsetting the feature - 
if we _did_ add the disable logic as badges had it, could we - in order to prevent accidentally unsetting a feature flag due to the input being disabled, rely on the current `$user` prop as a second condition here?
e.g. 
```php
'refer-friends' => in_array('refer-friends', $input['feature_flags']) || date_get($user, 'feature_flags.refer-friends', false),
```

This could very well fall under the category of 'too complicated' lol, but just a thought!

aaronschachter: I can give it a look! I will merge this for now to keep ⚽️ 🎢

PR #222, opened 2019-09-20
Editable feature flags

aaronschachter:
Fixes a bug in #221 to allow unsetting feature flags. I know when #220 introduced the field, it made it irremovable, but it'd be easiest to allow this for testing RAF, at least for now.

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -82,7 +82,7 @@ class UsersController extends Controller
                 'refer-friends' => in_array('refer-friends', $input['feature_flags']),
             ];
         } else {
-            $input['feature_flags'] = null;
+            $input['feature_flags'] = [];
         }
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -82,7 +82,7 @@ public function update(NorthstarUser $user, Request $request)
                 'refer-friends' => in_array('refer-friends', $input['feature_flags']),
             ];
         } else {
-            unset($input['feature_flags']);
+            $input['feature_flags'] = null;
```
mendelB: Is there any difference between this and `email_subscription_topics` above where we use an empty object?

aaronschachter: I thought it'd save an empty array to the user model, but it doesn't. I'll update so it's consistent.

PR #223, opened 2019-09-24
Re-organize for better UX & add support for sensitive fields.

DFurnes:
### What's this PR do?

This pull request adds support for "sensitive" fields to Aurora, which lets staffers view user profiles and optionally request to see personally-identifiable information when it's relevant to them (say, for contacting a scholarship winner).

This pull request is a little hefty, sorry! 😬 To make it less overwhelming, I rebased into logical commits & will give a little walkthrough explaining each one:

#### Dependency updates & tidying up:
To start, I've updated to the latest Gateway release (which uses Northstar's `v2/` endpoints) in 69b52c1 and did a little tidying up of the profile page in a7ba8e3.

#### Add support for displaying optional fields on profile:
I then added a new "revealer" (better name, idk?) component in 848d748, which is used to show and hide optional fields on the user's profile page (by adding/removing an `?include` query string to the page URL that's ~forwarded along to Northstar~ TBD!):

Since this is a traditional server-rendered application, showing or hiding an optional field requires a full page reload… which can be jarring (since you lose your scroll position). As a quick and dirty fix, I added [Turbolinks](https://github.com/turbolinks/turbolinks) in 7c85677 to make this smoother:

opt-fields

#### Update index page to use "preview" fields:
Similarly, I updated the index page to use "preview" fields instead of real emails & mobiles in 44cff77. I also added "last accessed at" since that seems like a reasonable thing to skim when searching! 🕵️‍♀️

Screen Shot 2019-09-24 at 11 19 20 AM

#### Remove ability for non-admins to edit users:
One downside to locking down access to sensitive fields more carefully is that we, well, need to lock them down more carefully! Since the "edit" page requires loading all sensitive fields on a user to populate the form, I've restricted it to admin users in 0ccaa8a. I don't think this should affect any standard use-cases, but worth considering a fancier solution if it does.

#### Human-readable countries & usernames in tab titles:
Finally, I made two small quality-of-life improvements: displaying the human-readable country name (rather than ISO code) on profiles in 5e56a91 and displaying the user's name in the page title in d9a915c (so it's easier to differentiate multiple tabs).

### Relevant Tickets

[#167596656](https://www.pivotaltracker.com/story/show/167596656)


DFurnes:
Note that this does not yet _actually_ forward `?include=` query strings to Northstar (because I forgot I need to expose that ability in Gateway, oops!). I'll make a corresponding PR there, and then updating this app to use that will be a nice and small change. 😌

mendelB:
> Note that this does not yet _actually_ forward `?include=` query strings to Northstar (because I forgot I need to expose that ability in Gateway, oops!). I'll make a corresponding PR there, and then updating this app to use that will be a nice and small change. 😌

So currently, it's just essentially mocking the `?include=` functionality to reveal/hide the built-in fields?

DFurnes:
Yep – for now, this only happens on the "front-end" in Aurora (so functionally it's "done" for users, but the underlying API request would still include the hidden fields). Northstar's "optional fields" functionality, while finished, is still behind a feature flag until we update all these apps.

I'll update Gateway to be able to forward along Aurora's `?include=` to Northstar, and then it'll be ready for when we flip that feature flag to enable this new behavior across the board.

resources/views/layout/master.blade.php
```diff
@@ -2,11 +2,15 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
-    <title>Aurora</title>
+    <title>
+        {{ isset($title) ? $title . ' | Aurora' : 'Aurora' }}
+    </title>
 
     <link rel="stylesheet" href="{{ elixir('app.css', 'dist') }}">
-    <script src="{{ asset('dist/modernizr.js') }}"></script>
```
mendelB: - kindly excusing my general ignorance of this script - what warranted this ✂️ ?

resources/views/users/partials/field.blade.php
```diff
@@ -0,0 +1,6 @@
+<dt>{{ $label or title_case($field) }}:</dt>
+@if ($user->{$field})
```
mendelB: Small fry but mostly curious -- could this just be `<dd>{{$user->{$field} or &mdash;}}</dd>` or is it different because we're accessing a $user attribute?

resources/views/users/partials/index-table.blade.php
```diff
@@ -1,19 +1,27 @@
...
+            <td class="table-cell"><a href="{{ route('users.show', [$user->id]) }}">{{ $user->display_name }}</a></td>
+            <td class="table-cell">
+                <code>{{ $user->email_preview }}</code>
+                @if ($user->email_preview && $user->mobile_preview)
+                    <span class="footnote"> and </span>
+                @endif
+                <code>{{ $user->mobile_preview }}</code>
+            </td>
+            <td class="table-cell footnote">
+                {{ $user->last_accessed_at ? $user->last_accessed_at->diffForHumans() : 'more than 2 years ago' }}
```
mendelB: `'more than 2 years ago'` -- assuming that's when we added the `last_accessed_at` field?

resources/views/users/partials/profile.blade.php
```diff
@@ -1,17 +1,30 @@
...
+    @include('users.partials.field', ['label' => 'First Name', 'field' => 'first_name'])
+    @include('users.partials.sensitive-field', ['label' => 'Last Name', 'field' => 'last_name', 'preview_field' => 'last_initial', 'preview_suffix' => '.'])
+    @include('users.partials.sensitive-field', ['field' => 'birthdate', 'preview_field' => 'age', 'preview_suffix' => ' years old'])
+    @include('users.partials.field', ['label' => 'Voter Registration Status', 'field' => 'voter_registration_status'])
+</div>
+<div class="profile-section">
+    <h4>Address:</h4>
+    @if ($user->addr_street1 || $user->addr_street2 || $user->addr_city || $user->addr_state || $user->addr_zip || $user->country)
+    <p>
+        @if (str_contains(request()->query('include'), 'addr_street1'))
```
mendelB: Why do we only check for `addr_street1` in the conditional? Is `addr_street2` assumed by default based on 1?

resources/views/users/partials/sensitive-field.blade.php
```diff
@@ -0,0 +1,8 @@
+<dt>{{ $label or title_case($field) }}:</dt>
+@if (str_contains(request()->query('include'), $field))
+    <dd class="revealed">{{ $user->{$field} }} {{ revealer($field) }}
+@elseif ($user->{$preview_field})
```
mendelB: Do _all_ sensitive fields have a `$preview_field` equivalent? (I guess otherwise, we'd potentially never be surfacing the revealer for the said field but maybe a moot point!

app/helpers.php
```diff
@@ -62,3 +62,31 @@ function scriptify($json = [], $store = 'STATE')
...
+ *
+ * @param  string $code
+ * @return string
+ */
+function country_name($code)
+{
+    $isoCodes = new \Sokil\IsoCodes\IsoCodesFactory();
+    $country = $isoCodes->getCountries()->getByAlpha2($code);
+
+    return $country ? $country->getName() : 'Unknown';
```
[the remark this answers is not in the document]
DFurnes: Oh, that's a great idea!! 👍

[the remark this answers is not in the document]
DFurnes: Updated in 98628d2.

resources/views/layout/master.blade.php
DFurnes: Ah sorry, should've noted! This is recommended [so that Turbolinks won't re-run scripts on every "in-app" page reload](https://github.com/turbolinks/turbolinks#working-with-script-elements) (since it more-or-less yanks the new `body` into the current page).

resources/views/users/partials/field.blade.php
DFurnes: It totally could, and that makes good sense to me. 👍 Swapped in cb24ae5.

app/Http/Controllers/UsersController.php
```diff
@@ -22,6 +22,8 @@ public function __construct(Northstar $northstar)
 
         $this->middleware('auth');
         $this->middleware('role:admin,staff,intern');
+
+        $this->middleware('role:admin', ['only' => ['edit', 'update']]);
```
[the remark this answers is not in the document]
DFurnes: I know that Hannah edits profiles from time to time when dealing with support tickets, but she's an admin so that'll still work. If it does end up disrupting anyone's workflow, I think we can either (temporarily) make them an admin or prioritize a better solution for this.

resources/views/layout/master.blade.php
mendelB: Wait sorry, so why do we need to get rid of `modernizr` completely for this?

DFurnes: Ohhh, whoops. I actually didn't mean to do that! 😳

DFurnes: Re-added this line in 9fff856.

resources/views/users/partials/sensitive-field.blade.php
DFurnes: Yeah, I think we'd need a different sort of interface. The address fields are actually a good example of this (and one of the reasons that they don't use the `sensitive-field` partial).

resources/views/users/partials/profile.blade.php
DFurnes: Yeah, but that was probably a little lazy of me! 😅 Updated in 8d51e86 to check for both.

resources/views/users/partials/index-table.blade.php
DFurnes: Yep! We added this in #591 so anyone who doesn't have it set has not visited the site since that was deployed (2 years and 3 months ago)... or I guess never (in the case of Chompy imports or SMS users).

It could maybe be clearer to just put `N/A` for those cases... what do you think?

mendelB: Hmmm ya know my  feeling is `N/A` is nice to account for the Chompy case, though I do still see the value in `more than 2 years ago` to easily spot legacy/inactive users 🤔 

BTW, would this field be updated if say an admin or a script touched the user instance? Or is this _only_ updated when they access the site?

DFurnes: Alright, yeah let's just go with `N/A` since it's a bit less confusing!

> BTW, would this field be updated if say an admin or a script touched the user instance? Or is this only updated when they access the site?

Only when the user accesses the site (via either logging in or returning to an existing web session & refreshing their access token).

DFurnes: Swapped to `N/A` in c51d99b.

PR #225, opened 2019-09-24
Send the `?include=` query string to Northstar.

DFurnes:
This is a quick follow-up to #223 that sends the `?include=` query string (which is used to conditionally display sensitive fields) along to Northstar, so that once we flip the feature-flag on this "optional fields" functionality this will all continue working!

~Blocked on https://github.com/DoSomething/gateway/pull/125 merge & release.~ Published as [v2.1.0](https://github.com/DoSomething/gateway/releases/tag/v2.1.0) and updated this app's dependencies in d27f972.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -14,7 +14,7 @@ function csv_query(string $key, array $default = []): array
     $query = request()->query($key);
 
     if (! $query) {
-       return $default;
+        return $default;
     }
 
     return explode(',', $query);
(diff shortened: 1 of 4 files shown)


app/helpers.php
```diff
@@ -2,6 +2,24 @@
...
 use Illuminate\Support\HtmlString;
 
+/**
+ * Read a given CSV-formatted query string.
+ *
+ * @param string $key
+ * @param string[] $default
+ * @return string[]
+ */
+function csv_query(string $key, array $default = []): array
```
mendelB: I legitimately did not realize CSV stood for comma-separated value until I read this in my head. What a day.

PR #226, opened 2019-10-09
Properly fetch optional fields for 'edit' form.

DFurnes:
This pull request includes a "quick fix" to the `users/{id}/edit` page, which was [not loading optional fields & so those fields would appear blank](https://dosomething.slack.com/archives/C02BBRN5A/p1570559704003600). Longer term, I'd like to spend a bit of time to get this working similarly to the "show" page (with 👁's), but this gets things back in working order!

References [#169023024](https://www.pivotaltracker.com/story/show/169023024)

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -17,7 +17,8 @@ $router->get('auth/login', 'Auth\AuthController@getLogin');
 $router->get('auth/logout', 'Auth\AuthController@getLogout');
 
 // Users
-$router->resource('users', 'UsersController', ['except' => ['create', 'store']]);
+$router->resource('users', 'UsersController', ['except' => ['create', 'edit', 'store']]);
+$router->get('users/{id}/edit', ['as' => 'users.edit', 'uses' => 'UsersController@edit']);
 $router->get('search', ['as' => 'user.search', 'uses' => 'UsersController@search']);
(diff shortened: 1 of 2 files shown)


routes/web.php
```diff
@@ -17,7 +17,8 @@
 $router->get('auth/logout', 'Auth\AuthController@getLogout');
 
 // Users
-$router->resource('users', 'UsersController', ['except' => ['create', 'store']]);
+$router->resource('users', 'UsersController', ['except' => ['create', 'edit', 'store']]);
+$router->get('users/{id}/edit', ['as' => 'users.edit', 'uses' => 'UsersController@edit']);
```
katiecrane: Did this have to be pulled out because of the `'as' => 'users.edit'` part?

DFurnes: I had to pull this out because the "resource" route definition uses route-model binding to automatically fetch the user from Northstar, and so doesn't give us an option to select which fields we want to see. By using a normal `{id}` parameter, we can then customize how we fetch it in `UsersController@edit`.

DFurnes: This is definitely not the cleanest approach, but gets things back in order!

katiecrane: ah that makes sense!

PR #233, opened 2020-02-25
Allow staff/admins to delete user accounts.

DFurnes:
### What's this PR do?

This pull request hooks up the "Delete Account" interface in Aurora to Northstar's new super-charged `DELETE v2/users/{id}` endpoint. This allows admins to delete a user from Northstar, Rogue, Gambit, and Customer.io on demand:

Screen Shot 2020-02-25 at 1 20 50 PM

Aurora will also now display a banner if an account has been queued for deletion:

Screen Shot 2020-02-25 at 1 21 28 PM

I've also fixed up the homepage copy to direct users to `#help-product` if they need permissions.

👀

I decided to hold off on allowing admins to queue/de-queue accounts for deletion, since I think if a user requests their account to be deleted we should be very careful about "undo-ing" that change.

### Relevant tickets

References [Pivotal #170953777](https://www.pivotaltracker.com/story/show/170953777).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.

### DIFF
--- a/resources/views/users/partials/danger-zone.blade.php
+++ b/resources/views/users/partials/danger-zone.blade.php
@@ -21,11 +21,10 @@
         {!! Form::open(['route' => ['users.destroy', $user->id], 'method' => 'DELETE']) !!}
         <div class="form-item">
             <label for="role" class="field-label">Delete Account</label>
-            <p class="footnote">This will <strong>permanently remove</strong> a user's account from Northstar.
-                This is the point of no return! Beware!</p>
+            <p class="footnote">This will <strong>permanently destroy</strong> this user's Northstar & Customer.io profiles, Rogue campaign activity, and Gambit conversation history.</p>
         </div>
         <div class="form-actions">
-            {{ Form::submit('Delete Now', [
+            {{ Form::submit('Delete Immediately', [
                 'class' => 'button -secondary -danger',
                 'data-confirm' => 'Are you sure you want to immediately & permanently destroy this user? THIS CANNOT BE UNDONE.'
             ]) }}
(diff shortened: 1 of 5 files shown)


resources/views/users/partials/danger-zone.blade.php
```diff
@@ -21,11 +21,13 @@
...
             <label for="role" class="field-label">Delete Account</label>
-            <p class="footnote">This will <strong>permanently remove</strong> a user's account from Northstar.
-                This is the point of no return! Beware!</p>
+            <p class="footnote">This will <strong>permanently destroy</strong> this user's Northstar & Customer.io profiles, Rogue campaign activity, and Gambit conversation history.</p>
         </div>
         <div class="form-actions">
-            {!! Form::submit('Delete User', ['class' => 'button -secondary -danger']) !!}
+            {{ Form::submit('Delete Immediately', [
+                'class' => 'button -secondary -danger',
+                'data-confirm' => 'Are you sure you want to immediately & permanently destroy this user? THIS CANNOT BE UNDONE.'
```
mendelB: lol this feels so scary! Would it be overkill to have one of those confirmation forms where you need to manually input e.g. the Northstar ID to confirm the deletion?

lindsayrmaher: ooooo i do think that would be a good idea as well

DFurnes: Yeah, it's meant to feel scary! 😈 I do like the idea of typing the user's ID to confirm that you're nuking the right user... created [#171472709](https://www.pivotaltracker.com/story/show/171472709) to come back and try that!

In the meantime, I added the user's display name to the confirmation prompt in 89492b5, which will hopefully help to reduce the chance of deleting the wrong account:

Screen Shot 2020-02-25 at 1 39 25 PM

PR #248, opened 2020-08-10
Upgrading to laravel56

weerd:
### What's this PR do?

This pull request upgrades Aurora from Laravel 5.5 to 5.6! 🆙 

👀 

### Relevant tickets

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.


config/trustedproxy.php
```diff
@@ -1,40 +1,50 @@
...
+     * directly to your server, use an array or a string separated by comma of IP addresses:
+     */
+    // 'proxies' => ['192.168.1.1'],
+    // 'proxies' => '192.168.1.1, 192.168.1.2',
+
+    /*
+     * Or, to trust all proxies that connect
+     * directly to your server, use a "*"
+     */
+    // 'proxies' => '*',
```
mendelB, line 31: Any reason we're commenting this all out if we're assigning this `proxies` value above?

config/trustedproxy.php
```diff
@@ -1,40 +1,50 @@
...
+     * @link https://symfony.com/doc/current/deployment/proxies.html
      */
-    'headers' => [
-        \Illuminate\Http\Request::HEADER_FORWARDED    => 'FORWARDED',
-        \Illuminate\Http\Request::HEADER_CLIENT_IP    => 'X_FORWARDED_FOR',
-        \Illuminate\Http\Request::HEADER_CLIENT_HOST  => 'X_FORWARDED_HOST',
-        \Illuminate\Http\Request::HEADER_CLIENT_PROTO => 'X_FORWARDED_PROTO',
-        \Illuminate\Http\Request::HEADER_CLIENT_PORT  => 'X_FORWARDED_PORT',
-    ],
+    'headers' => Illuminate\Http\Request::HEADER_X_FORWARDED_ALL,
```
mendelB, line 48: Why is this value different than the one in the `TrustProxies` middleware? (I think I get [what this is doing](https://laravel.com/docs/5.6/requests#configuring-trusted-proxies) but I'm not sure about the purpose of the config here vs in the middleware)

config/services.php
```diff
@@ -37,6 +37,12 @@
         'profile_url' => env('ROGUE_PROFILE_URL', 'https://activity.dosomething.org/users'),
     ],
 
+    'ses' => [
+        'key' => env('AWS_ACCESS_KEY_ID'),
+        'secret' => env('AWS_SECRET_ACCESS_KEY'),
+        'region' => env('AWS_DEFAULT_REGION', 'us-east-1'),
+    ],
```
mendelB, line 44: Do we use this Amazon SES service for email in Aurora? (I didn't think we sent emails from this application)

.env.example
```diff
@@ -1,6 +1,10 @@
 APP_ENV=local
-APP_DEBUG=true
 APP_KEY=
+APP_DEBUG=true
+APP_URL=http://aurora.test
+
+LOG_CHANNEL=daily
+LOG_SLACK_WEBHOOK_URL=
```
mendelB, line 7: Do we log to slack via this app anywhere? Or is this just in case we want to?

app/Http/Kernel.php
```diff
@@ -55,13 +55,14 @@ class Kernel extends HttpKernel
...
     protected $routeMiddleware = [
         'auth' => \Illuminate\Auth\Middleware\Authenticate::class,
-        'guest' => \Aurora\Http\Middleware\RedirectIfAuthenticated::class,
-        'role' => \Aurora\Http\Middleware\CheckRole::class,
-        'can' => \Illuminate\Auth\Middleware\Authorize::class,
-
-        'guard' => \DoSomething\Gateway\Server\Middleware\SetGuard::class,
         'auth.basic' => \Illuminate\Auth\Middleware\AuthenticateWithBasicAuth::class,
         'bindings' => \Illuminate\Routing\Middleware\SubstituteBindings::class,
+        'cache.headers' => \Illuminate\Http\Middleware\SetCacheHeaders::class,
```
mendelB, line 60: Looks like this and the `signed` are new middleware. Is that just par for the upgrade?

.env.example
weerd: We don't but adding it to be more consistent with Rogue. But I can remove if seems confusing.

config/services.php
weerd: Nope, but I wanted to update these to match the decisions we made for all AWS related env vars in Rogue to be consistent. Here's a [conversation](https://github.com/DoSomething/rogue/pull/1081/files#r464635014) about it, but there were others throughout the Rouge upgrade process.

config/trustedproxy.php
weerd: I mostly copy-pasted the config from the [package](https://github.com/fideloper/TrustedProxy/blob/master/config/trustedproxy.php) and appreciated the comments that helped clarify what the options are and what they do.

config/trustedproxy.php
weerd: Oh, hm... I copied these settings over from Rogue since both apps are on Heroku. Buuuut, you do bring up a good point. I wonder if we actually need to config file? I'm not 100% sure how the settings in the middleware coincide with the published config settings. 🤔  Typically the config is used to config the package, but since the middleware utilizes the package too I wonder how these settings get applied. @DFurnes any thoughts on this? Could mean an edit to Rogue as well.

app/Http/Kernel.php
weerd: Yep! Just wanted to make sure Aurora code was up-to-date with [5.6 framework code](https://github.com/laravel/laravel/blob/5.6/app/Http/Kernel.php) and these were added there!

config/trustedproxy.php
weerd: I suspect the config overrides whatever is declared in the middleware and we _should_ keep the middleware as it is in the framework and update the `config/trustedproxy.php` to be `HEADER_X_FORWARDED_AWS_ELB`.

mendelB: Hmm that does make sense to me! I think I understand you correctly -- given the option to simply assign this via config we don't even need our own custom middleware?

weerd: Well the middleware is [part of the framework](https://github.com/laravel/laravel/blob/master/app/Http/Middleware/TrustProxies.php) by default. I think publishing the config just gives us an option for overriding the default settings if you don't want to change what the default is in the framework code. But also not 100% sure on this given the Laravel docs clearly show editing the framework middleware code, haha 😅 ... so it's a bit confusing why have the publishable config from the `TrustedProxy` package?

mendelB: Yeah sorry, was referring to our own customized extended version in `App/Http/Middleware/TrustProxies.php`. I have the same confusion though about what should be going where, but like the config seems to make more sense!

mendelB: @weerd this [SO answer](https://stackoverflow.com/a/48595052/5657278) seems to recommend deleting the config in favor of the middleware override.

weerd: Ah! Excellent find @mendelB 🙌  

I'll make this change here, which definitely clarifies the situation and I'll make a cleanup ticket to update Rogue as well! Great 👁️  in questioning this!

DFurnes: That makes sense to configure in the middleware itself! Good find Mendel! 🎉

weerd: Thanks for the help team! 🤝

PR #253, opened 2020-08-13
Upgrading Laravel to 5.8

weerd:
### What's this PR do?

This pull request upgrades Aurora from Laravel 5.7 to 5.8! 🆙 

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,17 +6,17 @@ APP_URL=http://aurora.test
 LOG_CHANNEL=daily
 
 DB_CONNECTION=mysql
-DB_HOST=127.0.0.1
-DB_PORT=3306
-DB_DATABASE=aurora
-DB_USERNAME=homestead
-DB_PASSWORD=secret
+DATABASE_URL=mysql://homestead:secret@127.0.0.1:3306/aurora
+# @TODO Remove need for JAWSDB_MARIA_URL
+JAWSDB_MARIA_URL=mysql://homestead:secret@127.0.0.1:3306/aurora
 
 CACHE_DRIVER=file
 FILESYSTEM_DRIVER=public
 QUEUE_CONNECTION=sync
 SESSION_DRIVER=file
 SESSION_LIFETIME=120
+
+REDIS_URL=redis://h:null@127.0.0.1:6379
 
 NORTHSTAR_URL=http://northstar.test
 NORTHSTAR_CLIENT_ID=dev-oauth
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -14,12 +14,12 @@ class Kernel extends HttpKernel
      * @var array
      */
     protected $middleware = [
+        \Aurora\Http\Middleware\TrustProxies::class,
+        \Aurora\Http\Middleware\ForceHttps::class,
         \Aurora\Http\Middleware\CheckForMaintenanceMode::class,
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \Aurora\Http\Middleware\TrimStrings::class,
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
-        \Aurora\Http\Middleware\TrustProxies::class,
-        \Aurora\Http\Middleware\ForceHttps::class,
     ];
 
     /**
--- a/config/database.php
+++ b/config/database.php
@@ -1,5 +1,7 @@
 <?php
 
+use Illuminate\Support\Str;
+
 // Convert Heroku's `REDIS_URL` to standard `REDIS_HOST`, `REDIS_PORT`,
 // and `REDIS_PASSWORD` environment variables:
 if (env('REDIS_URL')) {
@@ -54,6 +56,7 @@ return [
 
         'sqlite' => [
             'driver' => 'sqlite',
+            'url' => env('DATABASE_URL'),
             'database' => env('DB_DATABASE', database_path('database.sqlite')),
             'prefix' => '',
             'foreign_key_constraints' => env('DB_FOREIGN_KEYS', true),
@@ -61,6 +64,7 @@ return [
 
         'mysql' => [
             'driver' => 'mysql',
+            'url' => env('DATABASE_URL'),
             'host' => env('DB_HOST', '127.0.0.1'),
             'port' => env('DB_PORT', '3306'),
             'database' => env('DB_DATABASE', 'forge'),
@@ -73,10 +77,14 @@ return [
             'prefix_indexes' => true,
             'strict' => false,
             'engine' => null,
+            'options' => extension_loaded('pdo_mysql') ? array_filter([
+                PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
+            ]) : [],
         ],
 
         'pgsql' => [
             'driver' => 'pgsql',
+            'url' => env('DATABASE_URL'),
             'host' => env('DB_HOST', '127.0.0.1'),
             'port' => env('DB_PORT', '5432'),
             'database' => env('DB_DATABASE', 'forge'),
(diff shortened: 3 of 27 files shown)


config/database.php
```diff
@@ -1,23 +1,6 @@
...
-
-// Convert Heroku's `JAWSDB_MARIA_URL` to standard `DB_*` variables:
-if (env('JAWSDB_MARIA_URL')) {
-    $url = parse_url(env('JAWSDB_MARIA_URL'));
-    putenv('DB_HOST='.$url['host']);
-    putenv('DB_PORT='.$url['port']);
-    putenv('DB_USERNAME='.$url['user']);
-    putenv('DB_PASSWORD='.$url['pass']);
-    putenv('DB_DATABASE='.str_replace('/', '', $url['path']));
-}
```
weerd, line 20: This was a bit odd. Similar to [Rogue](https://github.com/DoSomething/rogue/pull/1071/files?file-filters%5B%5D=.example&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.xml&file-filters%5B%5D=.yml&file-filters%5B%5D=No+extension&file-filters%5B%5D=dotfile#r461733624) edits, I was able to remove the need to parse the `REDIS_URL` manually since Laravel 5.8 now supports a [`REDIS_URL env variable`](https://github.com/laravel/laravel/pull/5037/) that it then uses to populate the individual items.

However, while the `DATABASE_URL` should have been supported as well I was experiencing some odd behavior. While it does parse it and when configuring the database it uses the URL provided values, if you try to grab the same values using the `config()` helper, it would return the default values.

Tried troubleshooting but couldn't figure it out. I'm going to punt this to the 6.x upgrade and see if it's still an issue in that version. It otherwise works, just not when using the `config()`.

.env.example
```diff
@@ -6,18 +6,18 @@ APP_URL=http://aurora.test
 LOG_CHANNEL=daily
 
 DB_CONNECTION=mysql
-DB_HOST=127.0.0.1
-DB_PORT=3306
-DB_DATABASE=aurora
-DB_USERNAME=homestead
-DB_PASSWORD=secret
+DATABASE_URL=mysql://homestead:secret@127.0.0.1:3306/aurora
+# @TODO Remove need for JAWSDB_MARIA_URL
```
weerd, line 10: Related to [this comment](https://github.com/DoSomething/aurora/pull/253/files#r470305759). Need to update Heroku env vars and rename `JAWSDB_MARIA_URL` to `DATABASE_URL`.

app/Http/Kernel.php
```diff
@@ -14,12 +14,12 @@ class Kernel extends HttpKernel
...
      */
     protected $middleware = [
+        \Aurora\Http\Middleware\TrustProxies::class,
+        \Aurora\Http\Middleware\ForceHttps::class,
         \Aurora\Http\Middleware\CheckForMaintenanceMode::class,
         \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \Aurora\Http\Middleware\TrimStrings::class,
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
-        \Aurora\Http\Middleware\TrustProxies::class,
-        \Aurora\Http\Middleware\ForceHttps::class,
```
katiecrane, line 22: Why did these move up to the top?

config/queue.php
```diff
@@ -45,15 +45,17 @@
             'driver' => 'beanstalkd',
             'host'   => 'localhost',
             'queue'  => 'default',
-            'ttr'    => 60,
+            'retry_after' => 90,
```
katiecrane, line 48: Should this still be 60?

database/factories/UserFactory.php
```diff
@@ -19,7 +20,7 @@
         'name' => $faker->name,
         'email' => $faker->unique()->safeEmail,
         'email_verified_at' => now(),
-        'password' => '$2y$10$TKh8H1.PfQx37YgCzwiKb.KjNyWgaHb9cbcoQgdIVFlYg7B77UdFm', // secret
+        'password' => '$2y$10$92IXUNpkjO0rOQ5byMi.Ye4oKoEa3Ro9llC/.og/at2.uheWG/igi', // password
```
katiecrane, line 23: Does this mean the seeded admin account with now have a password of `password` rather than `secret`?

resources/lang/en/passwords.php
```diff
@@ -13,7 +13,7 @@
     |
     */
 
-    'password' => 'Passwords must be at least six characters and match the confirmation.',
+    'password' => 'Passwords must be at least eight characters and match the confirmation.',
```
katiecrane, line 16: Looks like in Northstar we still only require 6 characters

config/database.php
katiecrane: Hmm I don't see either `DATABASE_URL` or `JAWSDB_MARIA_URL` being pulled by `config()` so I'm confused about this? Why do we need both?

weerd: I was using `config()` to test whether it was working correctly to parse the URL and distribute the values into the respective database config properties for `mysql`. It does show correctly with the redis config data, but not with the database.

I use both for now. On Heroku we specifically declare `JAWSDB_MARIA_URL` but I would like to change that to `DATABASE_URL`. Using both for now as we transition the env vars. Plan is to remove `JAWSDB_MARIA_URL` env variable.

app/Http/Kernel.php
weerd: `TrustedProxies` moved up in the framework code so I moved it as well, and `ForceHttps` is our custom one and should happen after `TrustedProxies` is all.

config/database.php
katiecrane: Ooooooo gotcha - didn't realize it was called something different on Heroku!

config/queue.php
weerd: Framework updated all to `90` and this is for [`beanstalkd`](https://github.com/laravel/laravel/blob/5.8/config/queue.php#L48) which we don't use anyway.

katiecrane: oh perfect haha that makes it easy!

database/factories/UserFactory.php
weerd: We don't actually use this factory for seeding. I believe it's all connected to Northstar or use a faked NS account. More framework related changes that don't directly affect this app.

katiecrane: OH duh this is Aurora! definitely thought I was in Northstar there for a minute

resources/lang/en/passwords.php
weerd: Might have been a conscious decision from Dave to not change it on NS end. Aurora is connected to NS so this doesn't actually apply anything to this app, just more framework related changes!

katiecrane: ooo so we don't use this file at all actually?

katiecrane: I wasn't sure if there was double validation going on somehow (which would have been confusing!)

weerd: Nope! This would trigger if we used the Laravel specific Auth for creating accounts, but we direct everything to NS.

config/database.php
weerd: Ah yep! I could have been clearer about that, lol. It was late and I was troubleshooting the weirdness for a while 😴 

Aiming to universalize all the env vars on Heroku so they're same across apps. Mostly handled by Terraform, but this app I don't think is Terraform controlled at the moment.

resources/lang/en/passwords.php
weerd: > I wasn't sure if there was double validation going on somehow (which would have been confusing!)

Oh ya, good question. I don't believe so since it'd need to go through the validation flow on Aurora and not on NS. I can double-check though on the next upgrade to confirm!

PR #254, opened 2020-08-14
Upgrading to Laravel 6.x

weerd:
### What's this PR do?

This pull request upgrades Aurora from Laravel 5.8 to 6.x! 🆙 

👀 

### Relevant tickets

### Checklist


resources/lang/en/passwords.php
```diff
@@ -13,9 +13,9 @@
     |
     */
 
-    'password' => 'Passwords must be at least eight characters and match the confirmation.',
```
mendelB, line 16: Why are we removing this line?

phpunit.xml
```diff
@@ -1,13 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit backupGlobals="false"
-         backupStaticAttributes="false"
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
```
mendelB, line 3: This whole file is a magic spell to me -- what's this update doing? (and what are those new servers we're adding below?)

config/queue.php
```diff
@@ -80,6 +80,7 @@
     */
 
     'failed' => [
+        'driver' => env('QUEUE_FAILED_DRIVER', 'database'),
```
mendelB, line 83: Is this for [this](https://laravel.com/docs/6.x/upgrade#sqs-environment-variables)?

config/auth.php
```diff
@@ -101,10 +101,23 @@
     'passwords' => [
         'users' => [
             'provider' => 'users',
-            'email' => 'auth.emails.password',
             'table' => 'password_resets',
             'expire' => 60,
+            'throttle' => 60,
```
mendelB, line 106: Do we even have auth in-house in this app? or are the changes in this file just following by example from other apps config?

app/Http/Kernel.php
```diff
@@ -42,7 +42,7 @@ class Kernel extends HttpKernel
         'api' => [
             'guard:api',
             'throttle:60,1',
-            'bindings',
+            \Illuminate\Routing\Middleware\SubstituteBindings::class,
```
mendelB, line 45: What's this doing here?

app/Http/Kernel.php
```diff
@@ -61,6 +61,7 @@ class Kernel extends HttpKernel
         'can' => \Illuminate\Auth\Middleware\Authorize::class,
         'guard' => \DoSomething\Gateway\Server\Middleware\SetGuard::class,
         'guest' => \Aurora\Http\Middleware\RedirectIfAuthenticated::class,
+        'password.confirm' => \Illuminate\Auth\Middleware\RequirePassword::class,
```
mendelB, line 64: I had a related question about auth generally below in `config/auth.php` that I think applies here too.

resources/lang/en/passwords.php
weerd: It was removed from the framework too. Explanation here: https://github.com/laravel/framework/pull/29480

phpunit.xml
weerd: I'm less familiar with these `xmlns` and `xsi` lines, but in like [PHP Unit 7.5 or 8](https://phpunit.readthedocs.io/en/9.2/configuration.html#appendixes-configuration-phpunit-php-server) they changed it to `<server />` instead of `<env />` for the items below. 

I think the decision was to write these env variables to the `$_SERVER` global instead of the `$_ENV`; related to this change in Laravel 5.8: https://github.com/laravel/framework/pull/27462

config/auth.php
weerd: Ya. After I do all the upgrade guide changes, I go and do a "github.com/compare" between previous current Laravel version project is on and then the next version it is being upgraded too. I scroll through and then make any/all changes shown in the less impacted files so that in the future a dev making further upgrades isn't surprised be growing deviations in the files and wondering what's up.

I think it is easier to grok the changes in an upgrade when there are less differences in the files, so my aim is always to bring all the file changes (even if change is small or doesn't necessarily apply to this project) as close to the latest upgraded version to minimize confusion.

This also really helps in making _our_ changes to framework files stand out so it is more obvious when comparing that we changed something versus something that was changed in the framework a number of versions back and we just never updated (this happened to me a few times during the process).

config/queue.php
weerd: I think it's for any failed queues, not just SQS; just a change the framework made.

app/Http/Kernel.php
weerd: We don't use it, but the framework made this change and I wanted to make sure the code in the project was up to speed with 6.x.

app/Http/Kernel.php
weerd: I think the same it was doing before but instead using the string class name instead of an alias `bindings` to be more explicit.

config/auth.php
mendelB: Super smart -- thanks for explaining your process! (Bc I love handing out extra work -- would it be worth codifying this flow in communal docs as a 'framework upgrade guide'?)

mendelB:
Thanks for all the explanations @weerd! 🙌

config/auth.php
weerd: If other agree with this process sure! It's something I specifically do but isn't necessarily the approach taken by others or expected. But if there's consensus on this being the approach then sure!

I can write something up for the communal docs if I have time before parental leave AND the team goes for this approach!

PR #266, opened 2020-10-16
Link To User's Club in Rogue

mendelB:
### What's this PR do?

This pull request links to a user's club in Rogue if available.

https://github.com/DoSomething/aurora/commit/3d1c17a023dc5f6989b77bb355c8c6038867d7c4 replaces the `ROGUE_PROFILE_URL` with a `ROGUE_URL`, and instead inlines the profile path directly. We seem to only use this in one spot, so the benefit of the general `ROGUE_URL` seems like a good tradeoff. (I opted not to maintain the additional `ROGUE_PROFILE_URL` since that seemed like overkill).

### How should this be reviewed?`
I'd love feedback on if this approach re env variables makes sense.

And if the actual link implementation makes sense, this could be neat to extract into a partial and apply for the users school ID field as well.

This should make the admin experience a lot more convenient when dealing with user clubs.

### Relevant tickets

References [Pivotal #174586698](https://www.pivotaltracker.com/story/show/174586698).

### Checklist

Screen Shot 2020-10-16 at 3 19 46 PM

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,7 @@ FASTLY_API_KEY=
 FASTLY_TABLE_REDIRECTS=
 FASTLY_TABLE_REDIRECT_TYPES=
 
+ROGUE_URL=https://rogue.test
+
 CUSTOMER_IO_PROFILE_URL=https://fly.customer.io/env/63547/people
-ROGUE_PROFILE_URL=https://rogue.test/users
 GAMBIT_PROFILE_URL=https://gambit-admin-staging.herokuapp.com/users
--- a/resources/views/users/show.blade.php
+++ b/resources/views/users/show.blade.php
@@ -79,7 +79,7 @@
                         <a href="{{ config('services.gambit.profile_url') }}/{{ $user->id }}">Gambit</a>
                     </li>
                     <li>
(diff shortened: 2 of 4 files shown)


.env.example
```diff
@@ -28,6 +28,7 @@ FASTLY_API_KEY=
 FASTLY_TABLE_REDIRECTS=
 FASTLY_TABLE_REDIRECT_TYPES=
 
+ROGUE_URL=https://rogue.test
```
lindsayrmaher, line 31: I definitely think this approach makes sense!

resources/views/users/show.blade.php
```diff
@@ -79,7 +79,7 @@
                         <a href="{{ config('services.gambit.profile_url') }}/{{ $user->id }}">Gambit</a>
                     </li>
                     <li>
-                        <a href="{{ config('services.rogue.profile_url') }}/{{ $user->id }}">Rogue</a>
+                        <a href="{{ config('services.rogue.url') }}/users/{{ $user->id }}">Rogue</a>
```
lindsayrmaher, line 82: I'm a fan of only storing the the base URL as an env and updating the path as needed 👍

lindsayrmaher (approved):
This approach / implementation makes sense to me!

